PR #19, opened 2017-07-25
Remove list editor and add menus to the tree editor

martinRenou:


martinRenou:
This PR removes a lot of tests and code which will be used later in a different file, I will open an other PR for the tree editor menu without the deleting of the list editor

PR #27, opened 2017-07-28
Use List(Instance()) instead of List()

### LINKED ISSUE #22: Use List(Instance(...)) instead of List(...)
In `WorkflowSettings` and `WorkflowModelView` we should use 
```python
List(Instance(...))
```

instead of 
```python
List(...)
```

martinRenou:
Closes #22

### DIFF
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -2,6 +2,8 @@ from traits.api import List, Instance
 
 from pyface.tasks.api import Task, TaskLayout, PaneItem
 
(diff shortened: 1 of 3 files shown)


force_wfmanager/wfmanager_task.py
```diff
@@ -10,9 +10,13 @@ class WfManagerTask(Task):
     id = 'force_wfmanager.wfmanager_task'
     name = 'Workflow Manager'
 
-    mco_bundles = List
-    data_source_bundles = List
-    kpi_calculator_bundles = List
+    mco_bundles = List(Instance('force_bdss.mco.i_mco_bundle.IMCOBundle'))
```
stefanoborini: import from the API

PR #29, opened 2017-07-28
Add save button which open a dialog for saving the workflow

martinRenou:


force_wfmanager/left_side_pane/workflow_settings.py
```diff
@@ -238,10 +241,17 @@ class WorkflowSettings(TraitsDockPane):
 
     workflow = Instance(WorkflowModelView)
 
+    save_button = Button("Save")
+
+    workflow_valid = Bool(False)
```
stefanoborini: just remove it for now.

PR #32, opened 2017-07-31
Adapts the MCO parameter factories according to the BDSS

martinRenou:


force_wfmanager/left_side_pane/workflow_settings.py
```diff
@@ -35,9 +34,11 @@ def new_mco_handler(self, editor, object):
 
     def new_parameter_handler(self, editor, object):
         """ Opens a dialog for creating a parameter """
+        available_factories = \
+            editor.object.workflow_model.mco.bundle.parameter_factories()
```
stefanoborini: This is too long. divide it into temporary intermediates.

stefanoborini: I think you can for example extract the model as a Property returning editor.object.workflow_model

PR #34, opened 2017-07-31
Add load Workflow menu option and implementation

martinRenou:


force_wfmanager/wfmanager_task.py
```diff
@@ -56,6 +62,16 @@ def save_workflow(self):
             with open(dialog.path, 'wr') as output:
                 writer.write(self.workflow, output)
 
+    def load_workflow(self):
+        """ Shows a dialog to load a workflow file """
+        dialog = FileDialog(action="open")
+        result = dialog.open()
+
+        if result is OK:
```
stefanoborini: Handle the exceptions coming from the workflow reader.

PR #38, opened 2017-08-01
Rename Workflow variables

martinRenou:
Rename Workflow variables so that it is easier to understand which is the workflow model view object and the workflow model object (`workflow_mv` and `workflow_m`)

I also made the workflow model view private inside the `WorkflowSettings` class

### DIFF
--- a/force_wfmanager/left_side_pane/workflow_settings.py
+++ b/force_wfmanager/left_side_pane/workflow_settings.py
@@ -29,14 +29,14 @@ class WorkflowHandler(Handler):
     def new_mco_handler(self, editor, object):
         """ Opens a dialog for creating a MCO """
         modal = NewEntityModal(
-            workflow=editor.object.workflow,
+            workflow_mv=editor.object._workflow_mv,
             available_factories=editor.object.available_mco_factories)
         modal.configure_traits()
 
     def new_parameter_handler(self, editor, object):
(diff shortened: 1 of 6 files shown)


force_wfmanager/left_side_pane/workflow_settings.py
```diff
@@ -237,28 +237,30 @@ class WorkflowSettings(TraitsDockPane):
     available_kpi_calculator_factories = List(Instance(
         BaseKPICalculatorBundle))
 
-    workflow = Instance(WorkflowModelView)
+    #: The workflow model view
+    _workflow_mv = Instance(WorkflowModelView)
 
-    workflow_model = Instance(Workflow)
+    #: The workflow model
+    workflow_m = Instance(Workflow)
```
martinRenou: allow_none=False

force_wfmanager/left_side_pane/workflow_settings.py
```diff
@@ -237,28 +237,30 @@ class WorkflowSettings(TraitsDockPane):
     available_kpi_calculator_factories = List(Instance(
         BaseKPICalculatorBundle))
 
-    workflow = Instance(WorkflowModelView)
+    #: The workflow model view
+    _workflow_mv = Instance(WorkflowModelView)
```
martinRenou: allow_none = False

PR #41, opened 2017-08-02
Create Analysis model

martinRenou:


force_wfmanager/central_pane/analysis.py
```diff
@@ -1,9 +1,52 @@
...
+class Result(HasStrictTraits):
+    name = Str()
+    value = Any()
+
+
+class EvaluationStep(HasStrictTraits):
+    _step = 0
+
+    @classmethod
+    def init_step_count(cls):
```
stefanoborini: I don't see the point of this. Just have an empty model. Actually, you don't even need the model at this stage. Just two empty panes.

martinRenou: This method will be useful when restarting the computation, you want to clear the result panes and restart the count of evaluation steps. I don't understand your comment about not needing the model, it is exactly the point of this PR to create the model.

martinRenou: And in a next PR I want to create the Table view, and in yet another PR I want to create a plot

PR #50, opened 2017-08-07
Create model views

martinRenou:
Create ModelView classes for the MCO parameters, data sources and KPI calculators.
This will help for the integration of the slots, because then the slots will be part of the DataSourceModelView and the KPICalculatorModelView.

### DIFF
--- a/force_wfmanager/left_side_pane/mco_parameter_model_view.py
+++ b/force_wfmanager/left_side_pane/mco_parameter_model_view.py
@@ -11,7 +11,10 @@ class MCOParameterModelView(ModelView):
     #: MCO parameter model
     model = Instance(BaseMCOParameter, allow_none=False)
 
-    #: The human readable name of the MCO parameter
+    #: The human readable name of the MCO parameter class
+    label = Str()
+
+    #: The user defined name of the MCO parameter
     name = Str()
 
     #: The type of the MCO parameter
@@ -24,8 +27,11 @@ class MCOParameterModelView(ModelView):
         kind="subpanel",
     )
 
-    def _name_default(self):
+    def _label_default(self):
         return get_factory_name(self.model.factory)
+
+    def _name_default(self):
+        return self.model.name
 
     def _type_default(self):
         return self.model.type
--- a/force_wfmanager/left_side_pane/mco_parameter_model_view.py
+++ b/force_wfmanager/left_side_pane/mco_parameter_model_view.py
@@ -1,4 +1,4 @@
-from traits.api import Instance, Str, on_trait_change
+from traits.api import Instance, Str
 
 from traitsui.api import View, Item, ModelView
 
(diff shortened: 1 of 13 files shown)


force_wfmanager/left_side_pane/mco_parameter_model_view.py
```diff
@@ -0,0 +1,45 @@
...
+    #: Base view for the MCO parameter
+    traits_view = View(
+        Item(name="name"),
+        Item(name="type"),
+        kind="subpanel",
+    )
+
+    @on_trait_change('name')
+    def update_name(self):
+        self.model.name = self.name
```
stefanoborini: do we need to do this transport manually, or is there a trick we don't know to do it automatically?

force_wfmanager/left_side_pane/mco_parameter_model_view.py
```diff
@@ -0,0 +1,45 @@
...
+from traitsui.api import View, Item, ModelView
+
+from force_bdss.api import BaseMCOParameter
+
+from .view_utils import get_factory_name
+
+
+class MCOParameterModelView(ModelView):
+    #: MCO parameter model
+    model = Instance(BaseMCOParameter, allow_none=False)
```
stefanoborini: when you define a modelview, there's an entry called "object" that refers to your model automatically. You need to check the modelview api.

force_wfmanager/left_side_pane/workflow_model_view.py
```diff
@@ -63,17 +65,54 @@ def add_entity(self, entity):
...
+
     def _get_mco_representation(self):
         if self.model.mco is not None:
             return [MCOModelView(model=self.model.mco)]
         else:
             return []
 
     def _get_data_sources_representation(self):
-        return self.model.data_sources
+        return [DataSourceModelView(model=data_source)
```
stefanoborini: You might lose UI state if you throw away and recreate your modelview every time.

stefanoborini: move this to issues.

force_wfmanager/left_side_pane/workflow_settings.py
```diff
@@ -60,26 +59,11 @@ def new_kpi_calculator_handler(self, editor, object):
...
-        workflow_m = editor.object.workflow_m
-        workflow_m.data_sources.remove(object)
-
-    def delete_kpi_calculator_handler(self, editor, object):
-        """ Delete a KPI Calculator from the workflow """
-        workflow_m = editor.object.workflow_m
-        workflow_m.kpi_calculators.remove(object)
+    def delete_entity_handler(self, editor, object):
+        """ Delete an element from the workflow """
+        editor.object._workflow_mv.remove_entity(object.model)
```
stefanoborini: make this public.

PR #56, opened 2017-08-10
Add current_file attribute

martinRenou:
We now keep in memory which file is currently edited, so that the application doesn't open a new dialog every time the user clicks on save

### DIFF
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -65,7 +65,10 @@ class WfManagerTask(Task):
             current_file = self.current_file
         else:
             dialog = FileDialog(
-                action="save as", default_filename="workflow.json")
+                action="save as",
+                default_filename="workflow.json",
+                wildcard='JSON files (*.json)|*.json|'
+            )
             result = dialog.open()
 
             if result is not OK:
@@ -87,7 +90,10 @@ class WfManagerTask(Task):
 
     def load_workflow(self):
         """ Shows a dialog to load a workflow file """
-        dialog = FileDialog(action="open")
+        dialog = FileDialog(
+            action="open",
+            wildcard='JSON files (*.json)|*.json|'
+        )
         result = dialog.open()
 
         if result is OK:
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -73,9 +73,17 @@ class WfManagerTask(Task):
 
             current_file = dialog.path
 
-        with open(current_file, 'w') as output:
-            writer.write(self.workflow_m, output)
-            self.current_file = current_file
+        try:
+            with open(current_file, 'w') as output:
+                writer.write(self.workflow_m, output)
+                self.current_file = current_file
+        except IOError as e:
+            error(
+                None,
+                'Cannot save in the requested file:\n\n{}'.format(
+                    str(e)),
+                'Error when saving workflow'
+            )
 
     def load_workflow(self):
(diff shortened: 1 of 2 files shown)


force_wfmanager/wfmanager_task.py
```diff
@@ -27,6 +27,9 @@ class WfManagerTask(Task):
     #: Registry of the available factories
     factory_registry = Instance(FactoryRegistryPlugin)
 
+    #: Current workflow file on which the application is writing
+    current_file = Str()
```
stefanoborini: Unicode

force_wfmanager/wfmanager_task.py
```diff
@@ -54,13 +57,22 @@ def create_dock_panes(self):
...
+            with open(self.current_file, 'wr') as output:
+                writer.write(self.workflow_m, output)
+            return
+
         dialog = FileDialog(action="save as", default_filename="workflow.json")
         result = dialog.open()
 
         if result is OK:
-            writer = WorkflowWriter()
             with open(dialog.path, 'wr') as output:
```
stefanoborini: just w

force_wfmanager/wfmanager_task.py
```diff
@@ -54,13 +57,22 @@ def create_dock_panes(self):
 
     def save_workflow(self):
         """ Shows a dialog to save the workflow into a JSON file """
+        writer = WorkflowWriter()
+
+        # If the user already saved before or loaded a file, we overwrite this
+        # one
+        if len(self.current_file) != 0:
+            with open(self.current_file, 'wr') as output:
```
stefanoborini: w

force_wfmanager/wfmanager_task.py
```diff
@@ -54,13 +57,22 @@ def create_dock_panes(self):
...
+            return
+
         dialog = FileDialog(action="save as", default_filename="workflow.json")
         result = dialog.open()
 
         if result is OK:
-            writer = WorkflowWriter()
             with open(dialog.path, 'wr') as output:
                 writer.write(self.workflow_m, output)
+            self.current_file = dialog.path
```
stefanoborini: Conglomerate the two operations, but don't store the file in the current_file until you are sure that the saving is successful.

force_wfmanager/wfmanager_task.py
```diff
@@ -54,13 +57,25 @@ def create_dock_panes(self):
...
+            dialog = FileDialog(
+                action="save as", default_filename="workflow.json")
+            result = dialog.open()
+
+            if result is not OK:
+                return
+
+            current_file = dialog.path
+
+        with open(current_file, 'w') as output:
```
stefanoborini: You need to handle exception here. Otherwise the first time the file can't be opened you get no response at all, the user thinks everything has been saved, while in reality there's a big exception in the console he can't see.

test/test_wfmanager_task.py
```diff
@@ -109,7 +149,7 @@ def test_load_failure(self):
                 mock.patch(ERROR_PATH) as mock_error, \
                 mock.patch(WORKFLOW_READER_PATH) as mock_reader:
             mock_dialog.side_effect = mock_file_dialog
-            mock_file_open.side_effect = mock_open
+            mock_open.side_effect = mock_file_open
```
stefanoborini: I'm not sure this work. side effect means you are setting the return value. This is what you need http://www.voidspace.org.uk/python/mock/helpers.html#mock-open

martinRenou: Thanks

PR #79, opened 2017-08-14
Add Save as... action to the application

### LINKED ISSUE #74: Change save menu
We should have two actions in the main menu:
- Save
- Save as...

`Save` is disabled by default, and enable if you already saved a first time or working on a loaded workflow.
`Save as...` always opens a dialog for saving in a new file.

martinRenou:
Fixes #74 

The application has now "Save" and "Save as..." actions. The save action is disabled by default, and enabled only if a Workflow has already been loaded from a file, or the save as action has already been processed. The save as action can always be triggered, and will change the current file on which the application is currently writing.

### DIFF
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -53,7 +53,6 @@ class WfManagerTask(Task):
             name='Save Workflow',
             method='save_workflow',
             accelerator='Ctrl+S',
-            enabled_name='current_file',
         ),
         TaskAction(
             name='Save Workflow as...',
@@ -78,11 +77,13 @@ class WfManagerTask(Task):
         return [self.side_pane]
 
     def save_workflow(self):
-        """ Shows a dialog to save the workflow into a JSON file """
+        """ Saves the workflow into the currently used file. If there is no
+        current file, it shows a dialog """
         if len(self.current_file) == 0:
-            return
+            return self.save_workflow_as()
 
-        self._write_workflow(self.current_file)
+        if self._write_workflow(self.current_file) is False:
+            self.current_file = ''
 
     def save_workflow_as(self):
         """ Shows a dialog to save the workflow into a JSON file """
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -103,7 +103,19 @@ class WfManagerTask(Task):
 
     def _write_workflow(self, file_path):
         """ Creates a JSON file in the file_path and write the workflow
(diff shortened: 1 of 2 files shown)


force_wfmanager/wfmanager_task.py
```diff
@@ -73,36 +79,57 @@ def create_dock_panes(self):
...
+            wildcard='JSON files (*.json)|*.json|'
+        )
+        result = dialog.open()
+
+        if result is not OK:
+            return
+
+        current_file = dialog.path
+
+        if self._write_workflow(current_file) is True:
```
stefanoborini: ==

force_wfmanager/wfmanager_task.py
```diff
@@ -73,36 +79,57 @@ def create_dock_panes(self):
...
-        if len(self.current_file) != 0:
-            current_file = self.current_file
-        else:
-            dialog = FileDialog(
-                action="save as",
-                default_filename="workflow.json",
-                wildcard='JSON files (*.json)|*.json|'
-            )
-            result = dialog.open()
+        self._write_workflow(self.current_file)
```
stefanoborini: reset the filename if it fails.

force_wfmanager/wfmanager_task.py
```diff
@@ -73,36 +79,57 @@ def create_dock_panes(self):
...
-                self.current_file = current_file
         except IOError as e:
             error(
                 None,
                 'Cannot save in the requested file:\n\n{}'.format(
                     str(e)),
                 'Error when saving workflow'
             )
+            return False
+        else:
```
stefanoborini: Handle generic exception as well.

force_wfmanager/wfmanager_task.py
```diff
@@ -73,36 +79,57 @@ def create_dock_panes(self):
...
+        Boolean:
+            True if it was a success to write in the file, False otherwise
+        """
+        writer = WorkflowWriter()
         try:
-            with open(current_file, 'w') as output:
+            with open(file_path, 'w') as output:
                 writer.write(self.workflow_m, output)
-                self.current_file = current_file
         except IOError as e:
```
stefanoborini: Do some logging.

force_wfmanager/wfmanager_task.py
martinRenou: Apparently, flake8 disagree with you:
```
./force_wfmanager/wfmanager_task.py:101:47: E712 comparison to True should be 'if cond is True:' or 'if cond:'```

PR #82, opened 2017-08-14
Future-based execution of the BDSS process

stefanoborini:
Reworks the BDSS execution to work in a concurrent future.


setup.py
```diff
@@ -22,21 +22,24 @@ def write_version_py():
 
 write_version_py()
 
+install_requires = [
```
martinRenou: Why did you need to export this from the setup arguments?

force_wfmanager/wfmanager_task.py
```diff
@@ -22,15 +22,7 @@
...
-@contextmanager
-def cleanup_garbage(tmpfile):
-    yield
-
-    try:
-        os.remove(tmpfile)
-    except OSError:
-        logging.exception("Could not delete the tmp file {}".format(tmpfile))
-        raise
+log = logging.getLogger(__name__)
```
martinRenou: It was already there

stefanoborini: Thanks!

PR #85, opened 2017-08-14
Add exit action in the menu bar

### LINKED ISSUE #80: Add exit action
An "Exit" button is added by default on Ubuntu, but apparently it is not the same behavior on MacOS. We should find a way to always have an "Exit" action in the menu with the "Ctrl+Q" accelerator.

martinRenou:
Adds an exit action in the menu bar
Fixes #80

### DIFF
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -63,6 +63,11 @@ class WfManagerTask(Task):
         ),
         SMenu(
             TaskAction(
+                name='Open Workflow...',
+                method='open_workflow',
+                accelerator='Ctrl+O',
+            ),
+            TaskAction(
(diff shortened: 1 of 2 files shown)


force_wfmanager/wfmanager_task.py
```diff
@@ -51,23 +52,34 @@ class WfManagerTask(Task):
...
+                name='Save Workflow',
+                method='save_workflow',
+                accelerator='Ctrl+S',
+            ),
+            TaskAction(
+                name='Save Workflow as...',
+                method='save_workflow_as',
+                accelerator='Shift+Ctrl+S',
+            ),
+            TaskAction(
```
stefanoborini: Rearrange position so that open comes first

PR #89, opened 2017-08-15
Made plot more robust and less exposed

stefanoborini:
- Slight changes in the plot to make it more robust to unexpected data in evaluation steps.
- Similar checks in the model
- some information made private.


force_wfmanager/central_pane/analysis_model.py
```diff
@@ -1,15 +1,72 @@
...
+                "with the number of value names {}.".format(
+                    evaluation_step, self.value_names))
+
+        self._evaluation_steps.append(evaluation_step)
+
+    def _get_selected_step_index(self):
+        return self._selected_step_index
+
+    def _set_selected_step_index(self, value):
+        if value is None:
```
martinRenou: We don't need this, do we ? It will be set at line 66

force_wfmanager/central_pane/plot.py
```diff
@@ -1,50 +1,45 @@
...
-
 from enable.api import Component, ComponentEditor
-
 from chaco.api import ArrayPlotData
 from chaco.api import Plot as ChacoPlot
 
 from .analysis_model import AnalysisModel
 
 
+log = logging.getLogger(__name__)
```
martinRenou: You never use it

force_wfmanager/central_pane/plot.py
```diff
@@ -97,15 +87,24 @@ def update_data_arrays(self):
...
         # If there is no data yet, or the data has been removed, make sure the
         # plot is updated accordingly and don't touch the data_arrays
-        if self.data_dim == 0:
+        if data_dim == 0:
             self._update_plot_data()
             return
 
         evaluation_steps = self.analysis_model.evaluation_steps
 
+        if data_dim != len(self._data_arrays):
```
martinRenou: Maybe add a comment for this case

PR #90, opened 2017-08-16
Add clear button in the central pane

### LINKED ISSUE #81: Add clear data button in the central pane
We should add a "Clear" button in the central pane to clear the current analysis model

martinRenou:
- It clears the model (empty the table and the plot)
- It is disabled when the computation is running

Fixes #81

### DIFF
--- a/force_wfmanager/central_pane/central_pane.py
+++ b/force_wfmanager/central_pane/central_pane.py
@@ -1,8 +1,8 @@
-from pyface.tasks.api import TraitsTaskPane
-
-from traits.api import Instance
+from traits.api import Instance, Bool, Button, on_trait_change
 
 from traitsui.api import View, Tabbed, VGroup, UItem
+
+from pyface.tasks.api import TraitsTaskPane
 
 from .analysis_model import AnalysisModel
 from .plot import Plot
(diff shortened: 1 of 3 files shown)


martinRenou:
I'm not completely sure if we need this button, as it is automatically cleared when running the computation again.

force_wfmanager/central_pane/central_pane.py
```diff
@@ -23,16 +23,28 @@ class CentralPane(TraitsTaskPane):
     #: The plot view
     plot = Instance(Plot)
 
-    view = View(Tabbed(
+    #: Enable or disable the contained entities.
+    #: Used when the computation is running
+    enabled = Bool(True)
+
+    #: Button used to clear the model
+    clear_button = Button('Clear')
```
stefanoborini: Ok, this is apparently a good idea, but a very dangerous one. The problem is that one click can delete all data accumulated over hours of evaluation. They will still be in the file, but still...

martinRenou: I completely agree with you, I wasn't sure if this button would be useful anyway because the run button does the clear when starting a new computation.

Which makes me think that when the user clicks on the run button and there is already result data, we should pop up a confirmation dialog.

martinRenou:
This clear button is useless as the result model will be cleared anyway when running an other computation. Also it is dangerous, the user would lost his results.

PR #95, opened 2017-08-16
Select the second component for the y axis of the plot by default

martinRenou:
For the plot, the second component of the evaluation steps is selected by default for the 'y' axis of the plot.


force_wfmanager/central_pane/plot.py
```diff
@@ -75,6 +75,13 @@ def __data_arrays_default(self):
     def _get__value_names(self):
         return self.analysis_model.value_names
 
+    @on_trait_change('_value_names')
```
stefanoborini: I have a suspect... that you could end up with an ordering problem.
both the comboboxes and this handler depend on value_names, but there's no guarantee that the comboboxes are updated first. It could be that this handler is called first, and it will try to set a value for y that is not there yet.

stefanoborini: One possible solution is that you invoke_later the change... but maybe there's a more elegant one. I'll think about it.

PR #99, opened 2017-08-17
Add comboboxes for selection of input slots

martinRenou:
Adds combo boxes for the selection of input slots.


force_wfmanager/left_side_pane/workflow_model_view.py
```diff
@@ -100,19 +101,57 @@ def remove_entity(self, entity):
...
+    def update_mco_parameters_names(self):
+        if len(self.mco_representation) != 0:
+            self.mco_parameters_names = \
+                self.mco_representation[0].mco_parameters_names
+        else:
+            self.mco_parameters_names = []
+
+    @on_trait_change('mco_parameters_names[]')
+    def update_data_sources_inputs(self):
+        for data_source_mv in self.data_sources_representation:
```
martinRenou: This scales badly, every time the user will type a new letter in a name of an MCO parameter, it will trigger this and be propagated everywhere.

martinRenou:
This first version doesn't work properly with the loading of the Workflow, I have to find out why

force_wfmanager/left_side_pane/mco_model_view.py
```diff
@@ -14,20 +15,18 @@ class MCOModelView(ModelView):
...
-    #: list of parameters in the MCO model
-    mco_parameters_representation = Property(
-        List(Instance(MCOParameterModelView)),
-        depends_on="model.parameters"
-    )
-
-    def _get_mco_parameters_representation(self):
-        return [MCOParameterModelView(model=parameter)
-                for parameter in self.model.parameters]
+    @on_trait_change('model.parameters[]')
```
stefanoborini: why did you change from property?

force_wfmanager/left_side_pane/variable_names_register.py
```diff
@@ -0,0 +1,63 @@
+from traits.api import HasStrictTraits, List, Instance, on_trait_change
+
+from force_bdss.api import Identifier
+
+from .mco_parameter_model_view import MCOParameterModelView
+from .data_source_model_view import DataSourceModelView
+from .kpi_calculator_model_view import KPICalculatorModelView
+
+
+class VariableNamesRegister(HasStrictTraits):
```
stefanoborini: Registry

stefanoborini: please add some comment on the purpose of the class and its interaction with the others because it's not immediate.

force_wfmanager/left_side_pane/mco_model_view.py
martinRenou: Because with a property it updates lazily. But I don't want to update the list of available parameters lazily, which relies on the change of this property.

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -0,0 +1,66 @@
+from traits.api import HasStrictTraits, List, Instance, on_trait_change
+
+from force_bdss.api import Identifier
+
+from .mco_parameter_model_view import MCOParameterModelView
+from .data_source_model_view import DataSourceModelView
+from .kpi_calculator_model_view import KPICalculatorModelView
+
+
+class VariableNamesRegistry(HasStrictTraits):
```
stefanoborini: Why would you need the modelviews in the variable names registry? this must be an exclusively model concern. The MVs should not be involved.

martinRenou: Because the `available_variables` attribute is part of the ModelViews, not of the Model. So in order to update the `available_variables`, I need the ModelViews.

martinRenou: And thinking about it, it does make sense. Because the `available_variables` attribute is only there to drive the possible values of the comboboxes, which is totally view related.

force_wfmanager/left_side_pane/evaluator_model_view.py
```diff
@@ -178,35 +200,67 @@ def _create_slots_tables(self):
...
-                          type=output_slot.type,
-                          model=self.model)
+                          type=output_slot.type)
             for index, output_slot in enumerate(output_slots)
         ]
+
+    @on_trait_change(
+        'variable_names_registry.data_source_available_variables[]')
+    def update_data_source_input_rows(self):
+        if not isinstance(self.model, BaseDataSourceModel):
```
stefanoborini: Why do you need to check for the type here?

force_wfmanager/left_side_pane/evaluator_model_view.py
```diff
@@ -178,35 +200,67 @@ def _create_slots_tables(self):
...
+            return
+
+        available_variables = \
+            self.variable_names_registry.data_source_available_variables
+        for input_slot_row in self.input_slots_representation:
+            input_slot_row.available_variables = available_variables
```
stefanoborini: Ok, so you are pushing the information in. I wonder if it would be a good idea to let the input_slot_rows fetch it instead, passing the registry to them.

force_wfmanager/left_side_pane/evaluator_model_view.py
```diff
@@ -178,35 +200,67 @@ def _create_slots_tables(self):
...
+
+    @on_trait_change(
+        'variable_names_registry.kpi_calculator_available_variables[]')
+    def update_kpi_calculator_input_rows(self):
+        if not isinstance(self.model, BaseKPICalculatorModel):
```
stefanoborini: I suspect you want these on the derived classes instead.

force_wfmanager/left_side_pane/tests/test_evaluator_model_view.py
```diff
@@ -61,7 +87,25 @@ def setUp(self):
...
         self.evaluator = factory.create_kpi_calculator()
 
-        self.evaluator_mv = EvaluatorModelView(model=self.model)
+        self.variable_names_registry = VariableNamesRegistry()
+        self.variable_names_registry.data_source_available_variables = \
+            ['P1']
+        self.variable_names_registry.kpi_calculator_available_variables = \
+            ['P1', 'P2', 'P3']
+
+        self.evaluator_mv = EvaluatorModelView(
```
stefanoborini: Isn't the evaluatormodelview the base class of the two? Why are you using the base class, instead of the specific ones?

force_wfmanager/left_side_pane/tests/test_evaluator_model_view.py
```diff
@@ -61,7 +87,25 @@ def setUp(self):
...
+            model=self.model,
+            variable_names_registry=self.variable_names_registry
+        )
+
+        data_source_factory = DataSourceFactory(mock.Mock(spec=Plugin))
+
+        data_source_model = data_source_factory.create_model()
+
+        self.data_source_mv = EvaluatorModelView(
```
stefanoborini: Which brings me to why do you need two?

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -0,0 +1,72 @@
...
+            data_sources_output_names.extend(data_source.output_slot_names)
+
+        # Removes empty strings from the variable names
+        while '' in data_sources_output_names:
+            data_sources_output_names.pop(data_sources_output_names.index(''))
+
+        self._data_sources_outputs = data_sources_output_names
+
+    @on_trait_change('_mco_parameters_names')
+    def update_available_variables(self):
```
stefanoborini: These two are fine for now, but I would add a comment that it must be a list of tuples instead.
When we'll have layers, it will be only one value we can monitor. This value will be the list of tuples, where you have variables available at each layer.

force_wfmanager/left_side_pane/evaluator_model_view.py
martinRenou: Because the `available_variables` won't be the same if it's a DataSource or a KPICalculator

force_wfmanager/left_side_pane/evaluator_model_view.py
martinRenou: Right, It would be better

force_wfmanager/left_side_pane/tests/test_evaluator_model_view.py
martinRenou: Absolutely, I will change this

force_wfmanager/left_side_pane/tests/test_evaluator_model_view.py
martinRenou: For coverage

force_wfmanager/left_side_pane/variable_names_registry.py
martinRenou: Ok

martinRenou: I've put a Note on top of the class

force_wfmanager/left_side_pane/evaluator_model_view.py
martinRenou: Then the `input_slot_row` would have to know if its model is a DataSource or a KPI (on which layer it is).

force_wfmanager/left_side_pane/evaluator_model_view.py
```diff
@@ -178,35 +189,45 @@ def _create_slots_tables(self):
...
         #: Initialize the output slots
-        self.output_slots_representation = []
         self.model.output_slot_names = len(output_slots)*['']
 
         self._fill_slot_rows(input_slots, output_slots)
 
     def _fill_slot_rows(self, input_slots, output_slots):
         """ Fill the tables rows according to input_slots and output_slots
         needed by the evaluator and the model slot values """
+        if isinstance(self.model, BaseDataSourceModel):
```
stefanoborini: still not delegated to

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -0,0 +1,77 @@
...
+    #: names and MCO parameters)
+    kpi_calculator_available_variables = List(Identifier)
+
+    #: List of MCO parameters
+    _mco_parameters_names = List(Identifier)
+
+    #: List of data sources outputs
+    _data_sources_outputs = List(Identifier)
+
+    def _data_source_available_variables_default(self):
```
stefanoborini: are redundant

PR #102, opened 2017-08-21
Add missing valid flag for the MCO parameter model view

martinRenou:


stefanoborini:
needs a test that would catch it missing...

Issue #58, opened 2017-08-10
Issue with slots and the "changes_slots" event in the models


martinRenou:
In this case scenario:
- create a KPIAdder model in the workflow
- set the input and output types of the KPI adder
- set a name for the three inputs and the output slots
- change the output type by editing the KPIAdder model, this will reinitialize the inputs and output names. Until then everything is ok.
- save the workflow in a json file
- load this json file
- the output name is back, but it should be empty (as when we saved the file)

I suspect that the output name in the table is initialized when changing the type, but it is not initialized in the model.

martinRenou:
This has been introduced with #53.
When calling `edit_traits` on the KPIAdder model, line 62 of `workflow_settings.py` with the option `kind=modal`, it creates a copy of the model, which is expected.
Now if you print the id and the value of the `output_slot_names` attribute before and after the `edit_traits` call you will have something like this:
```
before edit:
list id 140297824507256 
list value ['p2']

after edit:
list id 140297824167288 
list value ['p2']
```
So it made a copy as expected, didn't change the value of the list (because we only changed the output type, not the output value).
Now, because the `evaluator_model_view` reacts on the type change, it will empty the list of output names, we can put a print in the `_update_slots_tables` method, and this will be the output:
```
before edit:
list id 140297824507256 
list value ['p2']

when updating table:
list id 140297824166760
list value ['']

after edit:
list id 140297824167288 
list value ['p2']
```
So you can see something very weird happening there, the `evaluator_model_view` reacts during the `edit_traits` call, and it reacts on an other copy of the list. So at this exact moment, if you save the workflow, you will save the value of the `output_slots_names` which is saved in the model, which has not been reinitialized by the `evaluator_model_view` (['p2'] in this case). And the `evaluator_model_view` is pointing to the wrong list, and furthermore displaying the wrong list in the table.

Reacting on the `model.output_slot_names` in the `evaluator_model_view` to update the table doesn't fix this issue, because it overwrite the reinitialization of the names when changing the type of the slots.

The only solution I see is not using `kind=modal` with `edit_traits`. Because when closing the dialog (by pressing Ok OR Cancel), it overwrites the value of the `model.output_slot_names` with the values that were there before calling `edit_traits`. So the `model.output_slot_names` is not correctly reinitialized, and the `changes_slots` event on the model ends up having no repercussion.

stefanoborini:
Changed to livemodal to solve the issue. Fixed.

PR #112, opened 2017-08-22
Synchronize the slot tables with the model

martinRenou:
Synchronize the models with the slot tables.

This PR is a workaround for the issue #58, it doesn't fix the issue at all. But at least now the table shows to the user the correct current name of the slots.

This doesn't ensure that the type of the slot is synchronized with the model, but as "slots" method is not part of the model, we cannot know if the type changed.


stefanoborini:
needs tests.

PR #115, opened 2017-08-22
Disable the menu actions when the computation is running

martinRenou:
Disable the menus actions (save/save as.../load...) when the computation is running

### DIFF
--- a/force_wfmanager/wfmanager_task.py
+++ b/force_wfmanager/wfmanager_task.py
@@ -241,33 +241,32 @@ class WfManagerTask(Task):
                 return
 
         self._computation_running = True
-
-        for hook_manager in self._ui_hooks_managers:
-            try:
-                hook_manager.before_execution(self)
-            except Exception:
-                log.exception(
-                    "Failed before_execution hook "
-                    "for hook manager {}".format(
-                        hook_manager.__class__.__name__)
-                )
-
-        # Creates a temporary file containing the workflow
-        tmpfile_path = tempfile.mktemp()
         try:
+            for hook_manager in self._ui_hooks_managers:
+                try:
+                    hook_manager.before_execution(self)
+                except Exception:
+                    log.exception(
+                        "Failed before_execution hook "
+                        "for hook manager {}".format(
+                            hook_manager.__class__.__name__)
+                    )
+
+            # Creates a temporary file containing the workflow
+            tmpfile_path = tempfile.mktemp()
             with open(tmpfile_path, 'w') as output:
                 WorkflowWriter().write(self.workflow_m, output)
-        except Exception as e:
-            logging.exception("Unable to create temporary workflow file.")
-            error(None,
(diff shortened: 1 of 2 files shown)


force_wfmanager/wfmanager_task.py
```diff
@@ -54,6 +57,9 @@ class WfManagerTask(Task):
     #: ZeroMQ Server to receive information from the running BDSS
     _zmq_server = Instance(ZMQServer)
 
+    #: False when the ui must be disabled because of a running computation
+    _ui_enabled = Bool(True)
```
stefanoborini: just enabled.

force_wfmanager/wfmanager_task.py
```diff
@@ -68,16 +74,19 @@ class WfManagerTask(Task):
...
             TaskAction(
                 name='Save Workflow',
                 method='save_workflow',
+                enabled_name='_ui_enabled',
                 accelerator='Ctrl+S',
             ),
             TaskAction(
                 name='Save Workflow as...',
                 method='save_workflow_as',
+                enabled_name='_ui_enabled',
```
stefanoborini: isn't it enabled_when?

force_wfmanager/wfmanager_task.py
stefanoborini: actually, it's probably better to have a meaningful flag here. "computation_running". Because some things are actually enabled and can stay enabled while the evaluation is running, like the plot, or some menus (help)

force_wfmanager/wfmanager_task.py
martinRenou: Not for an action apparently, I have to define a name which refers to a Boolean trait

force_wfmanager/wfmanager_task.py
```diff
@@ -54,6 +57,12 @@ class WfManagerTask(Task):
     #: ZeroMQ Server to receive information from the running BDSS
     _zmq_server = Instance(ZMQServer)
 
+    #: Flag which says if the computation is running or not
+    _computation_running = Bool(False)
+
+    #: Flag which says if the menus should be disabled or not
+    _menu_enabled = Bool(True)
```
stefanoborini: why do you need an additional flag? Those menus only depend on computation running.

martinRenou: Because TaskActions are enabled/disabled with `enabled_name` which is the name of the boolean trait which define if the action is enabled or not.

martinRenou: I could have `_computation_not_running` but I don't like it

martinRenou: If would prefer an `enabled_when="not _computation_running"` but apparently it's not how TaskAction works

stefanoborini: never have flags with negating names. They are extremely confusing, especially when you end up double negating them for whatever reason.

PR #134, opened 2018-04-26
Fix for missing variable name at initialization

stefanoborini:
In some cases, the variable names were not set post loading a file. This is due to ordering of traits execution at initialisation of the class with the default data.

Not adding tests as this code will eventually disappear when we move to multilayering.


stefanoborini:
This code has done its part and with multilayers the issue is not there anymore

PR #140, opened 2018-05-10
Added windows decorations to the model visualisation

### LINKED ISSUE #183: Add window decorations to model
The model window has no decoration, like a group box around it. We need to add it.

stefanoborini:
Made the windows nicer by adding decoration boxes around them

Fixes #183

### DIFF
--- a/force_wfmanager/left_side_pane/new_entity_modal.py
+++ b/force_wfmanager/left_side_pane/new_entity_modal.py
@@ -217,14 +217,15 @@ class NewEntityModal(HasStrictTraits):
 
         # A default message when no model selected
         if self.selected_factory is None or self.current_model is None:
-            return html_string.format("No Model Selected", "")
+            return html_string.format("No Model Selected", "", "")
 
         model_name = self.selected_factory.get_name()
+        model_desc = self.selected_factory.get_description()
         view_info = self.view_structure()
 
         # Message for a model without editable traits
         if view_info == []:
-            return html_string.format(model_name, "")
+            return html_string.format(model_name, model_desc, "")
 
         # Remove traits in the view which are not editable
         view_info = [trait_name for trait_name in
@@ -249,7 +250,7 @@ class NewEntityModal(HasStrictTraits):
         body_str = ''.join([title_para.format(name, desc)
                            for name, desc in view_info])
 
-        return html_string.format(model_name, body_str)
+        return html_string.format(model_name, model_desc, body_str)
 
     def view_structure(self):
         """Return a list of editable traits in the order
@@ -259,7 +260,7 @@ class NewEntityModal(HasStrictTraits):
         #: A List containing the Groups/Items in this View
         main_group_contents = current_model_view.content.content
         #: A List of items from our view in the order they appear in the view.
-        #: This function does not do anything clever for unusual view layouts
+        #: This function does not do anything clever for unusual view layouts.
         main_group_items = _item_info_from_group(main_group_contents)
         return main_group_items
 
@@ -297,6 +298,7 @@ html_string = """
         </head>
         <body>
         <h1>{}</h1>
+        <p>{}</p>
         {}
         </body>
         </html>
--- a/force_wfmanager/left_side_pane/new_entity_modal.py
+++ b/force_wfmanager/left_side_pane/new_entity_modal.py
@@ -1,8 +1,9 @@
 from traits.api import (HasStrictTraits, Instance, List, Either,
-                        on_trait_change, Dict, Str, Property, HTML, Bool)
+                        on_trait_change, Dict, Str, Property, HTML, Bool,
+                        ReadOnly)
 from traitsui.api import (View, Handler, HSplit, Group, VGroup, UItem, Item,
                           InstanceEditor, OKCancelButtons, Menu,
-                          TreeEditor, TreeNode, HTMLEditor, TextEditor)
+                          TreeEditor, TreeNode, HTMLEditor)
 
 from envisage.plugin import Plugin
 
@@ -91,7 +92,7 @@ class NewEntityModal(HasStrictTraits):
 
     current_model_editable = Property(Bool, depends_on='current_model')
 
-    default_text = Str('No Configuration Options')
+    no_config_options_msg = ReadOnly(HTML)
 
     traits_view = View(
             HSplit(
@@ -106,9 +107,9 @@ class NewEntityModal(HasStrictTraits):
                               editor=InstanceEditor(),
                               visible_when='current_model_editable is True'
                               ),
-                        UItem('default_text',
+                        UItem('no_config_options_msg',
                               style='readonly',
-                              editor=TextEditor(),
+                              editor=HTMLEditor(),
                               visible_when='current_model_editable is False'),
                         visible_when='current_model is not None',
                         style="custom",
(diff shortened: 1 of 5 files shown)


stefanoborini:
There's one thing I am unsure about. When the model does not contain any trait, it appears as an ugly empty box. I am not sure if it's possible to hide it completely when there are no traits. Altenrnative is to leave it visible, but to stretch it.

stefanoborini:
Suggestion: either try to make it bigger, or have a tab with configuration options, containing the box you just introduced inside. That will then be expanded to have also the description.

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -96,21 +95,53 @@ class NewEntityModal(HasStrictTraits):
     #: Disable the OK button if no factory set
     OKCancelButtons[0].trait_set(enabled_when="selected_factory is not None")
 
+    model_description_HTML = Property(HTML, depends_on="current_model")
+
+    current_model_editable = Property(Bool, depends_on='current_model')
+
+    default_text = Str('No Configuration Options')
```
jjenthought: change to HTML

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+            return False
+
+        view_info = self.view_structure()
+
+        # Remove any non-visible traits
+        view_info = [trait_name for trait_name in
+                     view_info if trait_name in
+                     self.current_model.visible_traits()]
+
+        if view_info == []:
```
stefanoborini: minor thing. you can just return view_info != []

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+        """A check which indicates if 1. A view with at least
+        one item exists for this model and 2. That those items
+        are actually visible to the user"""
+        if self.current_model is None:
+            return False
+
+        view_info = self.view_structure()
+
+        # Remove any non-visible traits
+        view_info = [trait_name for trait_name in
```
stefanoborini: mostly stylistic as we normally split long list comprehensions. we put the for and the if on different lines as the first word, so it reads more or less like English.

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+
+        # Message for a model without editable traits
+        if view_info == []:
+            return html_string.format(model_name, model_desc, "")
+
+        # Remove traits in the view which are not editable
+        view_info = [trait_name for trait_name in
+                     view_info if trait_name in
+                     self.current_model.visible_traits()]
+
```
stefanoborini: uhm... this is a bit dodgy. You have a list of strings...

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+        # Retrieve descriptions from trait metadata
+        for i, trait_name in enumerate(view_info):
+            trait = self.current_model.trait(trait_name)
+            trait_desc = trait.desc
+
+            if trait_desc is not None:
+                view_info[i] = [trait_name, trait_desc]
```
stefanoborini: but then you change it to a list of lists, and modify it while iterating on it. I suggest using a different list.

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+        they appear in the view for the current model"""
+        #: The View of current_model
+        current_model_view = self.current_model.trait_view()
+
+        #: A List containing the Groups/Items in this View
+        main_group_contents = current_model_view.content.content
+
+        #: A List of items from our view in the order they appear in the view.
+        #: This function does not do anything clever for unusual view layouts.
+        main_group_items = _item_info_from_group(main_group_contents)
```
stefanoborini: you can just return the result of the function without the intermediate var

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+
+
+def _item_info_from_group(group_contents, item_info=None):
+    """Gets the item names from a list of groups (group_contents).
+    Returns a list of trait names corresponding to the items in the group.
+    """
+    if item_info is None:
+        item_info = []
+
+    for object in group_contents:
```
stefanoborini: use obj. object is a class name in the global scope, you don't want to overwrite global scope objects.

force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -153,3 +190,126 @@ def get_plugin_from_factory(self, factory):
...
+        if isinstance(object, Group):
+            item_info = _item_info_from_group(object.content, item_info)
+        # For an Item, add the item's name to item_info
+        elif isinstance(object, Item):
+            item_info.append(object.name)
+    return item_info
+
+
+# A generic HTML header and body with title and text
+html_string = """
```
stefanoborini: I would go for capitalised (constant like), so HTML_STRING. Actually, it's a template, so HTML_TEMPLATE, same for the title_para

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -32,7 +33,12 @@
 # Actions!
 new_mco_action = Action(name='New MCO...', action='new_mco')
 delete_mco_action = Action(name='Delete', action='delete_mco')
-edit_mco_action = Action(name='Edit...', action='edit_mco')
+edit_mco_action = Action(name='Edit...', action='edit_mco',
+                         enabled_when='handler.mv_editable(object)')
```
stefanoborini: in this case you are forced to use object, because it's the way it was designed (likely before the object class was introduced, some of our stuff is really ancient)

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -288,11 +318,46 @@ def delete_data_source(self, ui_info, object):
...
+        ModelEditDialog(model=object.model).edit_traits()
 
     def new_layer(self, ui_info, object):
         self.workflow_mv.add_execution_layer(ExecutionLayer())
 
     def delete_layer(self, ui_info, object):
         """ Delete an element from the workflow """
         self.workflow_mv.remove_execution_layer(object.model)
+
+    def mv_editable(self, modelview):
```
stefanoborini: this code has a lot in common with the other code in _get_current_model_editable. There's potential for factoring out the commonalities in a single function

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -288,11 +318,46 @@ def delete_data_source(self, ui_info, object):
...
+
+
+def _item_info_from_group(group_contents, item_info=None):
+    """Gets the item names from a list of groups (group_contents).
+    Returns a list of trait names corresponding to the items in the group.
+    """
+    if item_info is None:
+        item_info = []
+    for object in group_contents:
+        if isinstance(object, Group):
```
stefanoborini: obj

Issue #119, opened 2017-08-23
Introduce cli module for holding the run() method


stefanoborini:
As in BDSS, we need a cli module to keep the access point to the executable, just for a cleaner approach.

PR #149, opened 2018-06-06
gui module for force-wfmanager 

jjenthought:
A cli module containing the application launcher and tests which check if the application launched correctly. Addresses issue #119


force_wfmanager/cli/tests/test_execution.py
```diff
@@ -0,0 +1,31 @@
...
+
+        # available_commands is a list with elements of type: 'bytes'
+        for command in available_commands:
+            if command == b'force_wfmanager':
+                command_found = True
+        if command_found is False:
+            self.fail("Command \'force_wfmanager\' does not exist in the list of available bash commands")
+
+        # Does it execute correctly?
+        print("Close the workflow manager to complete the test!")
```
stefanoborini: do not put print statements in tests, your tests should have nothing on stdout/stderr.

force_wfmanager/cli/tests/test_execution.py
```diff
@@ -0,0 +1,31 @@
...
+import subprocess
+
+
+class TestExecution(unittest.TestCase):
+
+    def test_command_line_invocation(self):
+        """Test if the workflow manager can be loaded from the command line"""
+
+        # Does the command exist?
+        available_commands = subprocess.check_output(['/bin/bash', '-c', 'compgen -c'])
```
stefanoborini: this is not portable, but in any case you don't need it. The test will fail anyway if the command is not available.
Just invoke it

force_wfmanager/cli/tests/test_execution.py
stefanoborini: This will also hang with the automated CI. Tests should not require manual intervention. They should be fully automatic and report if it's ok or not.

stefanoborini: You have two strategies for this. Either you invoke the application with something that exits, like a --help or --version option (which we might need) or you kill the process after a given timeout and then take the timeout exception as a signal that everything is fine (because the application started)

force_wfmanager/cli/tests/test_execution.py
```diff
@@ -0,0 +1,31 @@
...
+            if command == b'force_wfmanager':
+                command_found = True
+        if command_found is False:
+            self.fail("Command \'force_wfmanager\' does not exist in the list of available bash commands")
+
+        # Does it execute correctly?
+        print("Close the workflow manager to complete the test!")
+        try:
+            subprocess.check_output(["force_wfmanager"])
+        except subprocess.CalledProcessError:
```
corranwebster: Rather than catching the exception and failing, just let the exception occur.  The test runner will let you know that something failed, and the traceback will be displayed with more useful information than just "couldn't load workflow manager".

force_wfmanager/cli/tests/test_execution.py
corranwebster: Additionally, the user won't see the text that is printed because most unit testing frameworks capture standard out and don't display it on the screen unless something goes wrong (or maybe in a very verbose mode).

setup.py
```diff
@@ -45,7 +45,7 @@ def write_version_py():
     package_data={'force_wfmanager.left_side_pane': 'icons/*'},
     entry_points={
         'gui_scripts': [
-            ('force_wfmanager = force_wfmanager.run:main')
+            ('force_wfmanager = force_wfmanager.cli.run:main')
```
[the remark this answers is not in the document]
stefanoborini: Ok, this addresses a question that @jjenthought raised about this specific issue. Good call. I didn't know that. We should certainly have gui then, instead of cli. Sorry @jjenthought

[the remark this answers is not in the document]
jjenthought: Thanks for the advice! I think I got too caught up with the idea of making things 'user-friendly' with the compgen stuff. The idea was that it would show if the program was broken (test no.2) or if they had shell problems (test no.1) or hadn't loaded the edm environment. 
However I forgot: users won't actually be running these tests so no.1 is pretty redundant! Stefano has also suggested some additional ways to test things, since no.2 requires user input it's not so well suited for automated testing!
Since this change needs to be updated to incorporate some new tests, I'll switch the name over as well

stefanoborini:
You have conflicts, please let me know if you need help in resolving them. If you've never done it before, it can be annoying.

Issue #151, opened 2018-06-06
Command line argument for program version 


jjenthought:
Add a command line argument for program version. This will help give something simple to test against in issue #119

PR #152, opened 2018-06-06
Added --version command line arg

jjenthought:
Relates to issue #151 and will be used for tests of new cli module in issue #119

### DIFF
--- a/force_wfmanager/run.py
+++ b/force_wfmanager/run.py
@@ -1,5 +1,5 @@
 import logging
-import argparse
+import click
 
 from envisage.core_plugin import CorePlugin
 from envisage.ui.tasks.tasks_plugin import TasksPlugin
@@ -16,17 +16,12 @@ from traits.api import push_exception_handler
 
 from .version import __version__
 
-
 push_exception_handler(lambda *args: None, reraise_exceptions=True)
 
 
+@click.command()
+@click.version_option(version=__version__)
 def main():
-
-    parser = argparse.ArgumentParser()
-    parser.add_argument('--version', action='version',
-                        version='force_wfmanager version: '
-                                '{}'.format(__version__))
-    parser.parse_known_args()
 
     logging.basicConfig(filename="force_wfmanager.log", filemode="w")
     log = logging.getLogger(__name__)
--- a/force_wfmanager/run.py
+++ b/force_wfmanager/run.py
@@ -1,4 +1,5 @@
 import logging
+import argparse
 
 from envisage.core_plugin import CorePlugin
 from envisage.ui.tasks.tasks_plugin import TasksPlugin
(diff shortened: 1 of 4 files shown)


force_wfmanager/run.py
```diff
@@ -19,7 +19,8 @@
 push_exception_handler(lambda *args: None, reraise_exceptions=True)
 
 
-@click.command()
+@click.command(context_settings=dict(ignore_unknown_options=True,
+                                     allow_extra_args=True))
```
stefanoborini: Why do you need them?

jjenthought: I was just trying it to see if it made the integration test work. The version with argparse passed when I let it ignore unexpected arguments so I thought that was the issue here. But clearly it's not the case, perhaps because click passes an exit code which the test doesn't expect?

stefanoborini: You can't invoke a click decorated function as a function anymore. Either extract the code in a separate _main() that acts like a function and test that one, or use the click CliRunner, but the first solution is probably what you need

force_wfmanager/run.py
```diff
@@ -12,10 +13,19 @@
...
 push_exception_handler(lambda *args: None, reraise_exceptions=True)
 
 
+@click.command()
+@click.version_option(version=__version__)
 def main():
+    _main()
+
+
+def _main():
```
corranwebster: Again, on the "naming things is hard" front, I think click expects the decorated function name to match the name of the command (unless explicitly overridden), so in this situation I think you might want something like:
```
@click_command(name='force-wfmanager')
@click.version_option(version=__version__)
def force_wfmanager():
    main()

def main():
    ...
```

corranwebster: Also, another opportunity for a "drive-by" improvement: give the functions docstrings, even if just one-liners!

jjenthought: Will do! For future stuff with more complex docstrings is there a particular style to follow?

corranwebster: Yes - Numpy style: see https://numpydoc.readthedocs.io/en/latest/format.html and http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

force_wfmanager/tests/test_click_run.py
```diff
@@ -0,0 +1,13 @@
+import unittest
+from click.testing import CliRunner
+
+from force_wfmanager.run import force_wfmanager
+
+
+class TestClickRun(unittest.TestCase):
+    def test_click_main(self):
+        """Test if the command force_wfmanager works"""
```
corranwebster: On gotcha here: unittest will print out the docstring of a test as the name of the test to output if one is present, eg. something like
```
Test if the command force_wfmanager works ... passed
```
instead of
```
test_click_main (TestClickRunner) ... passed
```
Which can make finding which function went wrong hard.  So normally I avoid docstrings in test methods and use a plain old comment if needed.

force_wfmanager/run.py
```diff
@@ -12,10 +13,22 @@
...
+
+from .version import __version__
+
 push_exception_handler(lambda *args: None, reraise_exceptions=True)
 
 
+@click.command()
+@click.version_option(version=__version__)
+def force_wfmanager():
+    """Launches the workflow manager, including arguments given
```
corranwebster: Second gotcha here: click will take the docstring and use it as the summary of the command when invoked with `--help` so you might want to write it with that in mind.  Something like "Launch the FORCE workflow manager application." is probably sufficient here.

corranwebster:
I think this looks good - on a diff this small don't worry about the codecov failure - it is one line of code and I think it will be fixed by the other PR.

stefanoborini:
Good for me too!

PR #159, opened 2018-06-07
Update dependencies to fix traitsui issue

### LINKED ISSUE #157: Upgrade traitsui to 6.0
There's a bug in the current version of traitsui that prevents us to do a few things with the tree editor. We need to upgrade to the most recent version.

Specifically, we cannot change the label in KPISpecificationModelView interactively.

jjenthought:
Fixes #157

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,7 +1,6 @@
 envisage == 4.6.0
 stevedore == 1.24.0
 futures == 3.1.1
-chaco == 4.7.1
 pyzmq==16.0.2
 six==1.10.0
 git+git://github.com/force-h2020/force-bdss.git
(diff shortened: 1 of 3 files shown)


requirements/requirements.txt
```diff
@@ -1,7 +1,7 @@
 envisage == 4.6.0
 stevedore == 1.24.0
-chaco == 4.6.1
 futures == 3.1.1
+chaco == 4.7.1
```
stefanoborini: we don't need it in requirements.txt anymore. requirements is installed by pip. if it's in edm, we already have chaco in the environment from edm.

It's a bit messy right now, but in principle we should have all dependencies in edm, and install nothing from pip (except for local things, such as force_bdss)

jjenthought: Ok, removed it from requriements.txt. We could also move six, pyzmq and envisage into edm_requirements as edm has up to date versions of those packages. Stevedore and futures don't seem to be up to date via edm right now though.

PR #171, opened 2018-06-12
Deletable data sources and extra KPI specification test

jjenthought:
Originally, deleting a datasource called a unwritten function in workflow_model_view. There is also an extra test for a named KPI, as we were only checking the default one before.
The KPI issue comes from the merge with master, but since this is such a small commit the single line of code not checked was causing the coverage checks to fail!

### DIFF
--- a/force_wfmanager/left_side_pane/tests/test_workflow_model_view.py
+++ b/force_wfmanager/left_side_pane/tests/test_workflow_model_view.py
@@ -8,15 +8,26 @@ from force_bdss.tests.probe_classes.notification_listener import \
     ProbeNotificationListenerFactory
 from force_bdss.tests.probe_classes.probe_extension_plugin import \
     ProbeExtensionPlugin
+
 from force_wfmanager.left_side_pane.workflow_model_view import \
     WorkflowModelView
-
+from force_wfmanager.left_side_pane.variable_names_registry import \
+    VariableNamesRegistry
+from force_bdss.tests.probe_classes.data_source import (ProbeDataSourceFactory,
+ProbeDataSource,ProbeDataSourceModel)
 
 # change
 class TestWorkflowModelView(unittest.TestCase):
     def setUp(self):
         self.wf_mv = WorkflowModelView(model=Workflow())
+        workflow = Workflow()
+        name_registry = VariableNamesRegistry(workflow)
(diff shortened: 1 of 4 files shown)


force_wfmanager/left_side_pane/tests/test_workflow_model_view.py
```diff
@@ -8,15 +8,27 @@
...
-# change
 class TestWorkflowModelView(unittest.TestCase):
     def setUp(self):
         self.wf_mv = WorkflowModelView(model=Workflow())
+        workflow = Workflow()
+        name_registry = VariableNamesRegistry(workflow)
+        self.wf_mv_name_registry = WorkflowModelView(
+            model=workflow, variable_names_registry=name_registry)
         self.plugin = ProbeExtensionPlugin()
+        self.datasource_models = [ProbeDataSourceModel(
```
corranwebster: Indentation is a bit funny on this line.

force_wfmanager/left_side_pane/tests/test_workflow_model_view.py
```diff
@@ -51,3 +63,18 @@ def test_remove_notification_listener(self):
         self.assertEqual(len(self.wf_mv.notification_listeners_mv), 1)
         self.wf_mv.remove_notification_listener(model)
         self.assertEqual(len(self.wf_mv.notification_listeners_mv), 0)
+
+    def test_remove_datasource(self):
+        self.wf_mv_name_registry.add_execution_layer(ExecutionLayer())
+        self.assertEqual(len(self.wf_mv_name_registry.
+                             execution_layers_mv[0].model.data_sources), 0)
+        self.wf_mv_name_registry.execution_layers_mv[0].\
```
corranwebster: Use a local variable here - if you are splitting a line on a `.` that's an indicator that you are chaining too deep.

If you do something like
```
execution_layer = self.wf_mv_name_registry.execution_layers_mv[0]
```
then I think that not only will this line be a lot cleaner, but so will subsequent lines.

jjenthought:
I probably could have split them up, the main goal was to have datasources you can delete but theres also a fix of a typo and a missed test of kpi’s I saw whilst working on things here. At the time I didn’t think they were big enough changes to open new issue/branches, but I’ll do that in future as it’s easier to track changes

PR #173, opened 2018-06-13
163 kpi name saves - partial

jjenthought:
Made kpi model track the name of the associated modelviewer. This prevents an error where nameless kpis would not be returned as output.
Also some slight changes to the MCO model viewer tests, as we now assign a name to the kpi model on initialisation. In a real run of the program this is passed as an argument to the kpi modelviewer, so the test was updated to reflect that.

This pr doesn't yet address the saving/loading behaviour, but I'd like to get it merged anyway as you can't run a computation in the workflow manager without this fix.


corranwebster:
I think I have a better solution for this.  Remove the `name` trait and related code from the modelview class, and change the view to be something like:
```
traits_view = View(
    Item('model.name', editor=EnumEditor(name='_combobox_values')),
    ...
```

PR #175, opened 2018-06-15
163 fixed names not loading for datasources

jjenthought:
When creating a workflow from a json file, the program uses this file to create the model - which should then automatically update the workflow model view. 
However with datasource names there was an issue where the Input row in the model view could not be assigned a name without first being assigned a list of available (name) variables. This then called functions designed to a) enusure the name was in available names  and b) enforce consistency between the modelview and model. The program essentially acted as if a user had selected the default name from the ui.
The fix was just to check a) elsewhere, in a way which did not trigger b) immediately.
Sorry for the long description, but it's such a small fix for a complicated (at least for me!) problem I felt it made sense to give some context.


corranwebster:
Hmmm looks like this was included in the other review I did.  When you have something like this where you have one PR depending on another then you should note in the PR that the other one should be merged first.

Issue #160, opened 2018-06-08
Make KPISpecificationModelView label update against the selected name


stefanoborini:
The label in the KPISpecificationModelView should follow the selected name, but at the moment we can't due to a bug in traits. We need to solve #157 first.

stefanoborini:
We now might be able to see it.

PR #176, opened 2018-06-15
160 update kpi mv label

jjenthought:
Makes workflow tree display track kpi names

### DIFF
--- a/force_wfmanager/left_side_pane/data_source_model_view.py
+++ b/force_wfmanager/left_side_pane/data_source_model_view.py
@@ -52,8 +52,17 @@ class InputSlotRow(TableRow):
     @on_trait_change('available_variables')
     def update_combobox_values(self):
         self._combobox_values = [''] + self.available_variables
-        self.name = ('' if self.name not in self.available_variables
-                     else self.name)
+
+        # Check for default name prevents spurious re-assignment when name is
+        # already ''. This causes problems when initialising a row from a json
+        # file, as we cannot change name without avail_vars. However, if this
+        # re-assignment occurs the corresponding model.input_slot_info name is
+        # also changed
+
+        if self.name != '':
+            self.name = ('' if self.name not in self.available_variables
+                         else self.name)
+
 
     def __combobox_values_default(self):
         return [''] + self.available_variables
--- a/force_wfmanager/left_side_pane/data_source_model_view.py
+++ b/force_wfmanager/left_side_pane/data_source_model_view.py
@@ -52,17 +52,8 @@ class InputSlotRow(TableRow):
     @on_trait_change('available_variables')
     def update_combobox_values(self):
         self._combobox_values = [''] + self.available_variables
-
-        # Check for default name prevents spurious re-assignment when name is
-        # already ''. This causes problems when initialising a row from a json
-        # file, as we cannot change name without avail_vars. However, if this
-        # re-assignment occurs the corresponding model.input_slot_info name is
-        # also changed
-
-        if self.name != '':
-            self.name = ('' if self.name not in self.available_variables
-                         else self.name)
-
+        self.name = ('' if self.name not in self.available_variables
+                     else self.name)
 
     def __combobox_values_default(self):
         return [''] + self.available_variables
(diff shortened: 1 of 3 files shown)


corranwebster:
It looks to me as if `label` should be a `Property` trait that `depends_on='model.name'`.

jjenthought:
@corranwebster Okay the changes seem to have gone through all right. If it looks good to you I'll merge it

force_wfmanager/left_side_pane/kpi_specification_model_view.py
```diff
@@ -40,19 +40,15 @@ def __init__(self, model, variable_names_registry, **kwargs):
...
-    def _label_default(self):
-        return _get_label(self.model)
-
     @on_trait_change('variable_names_registry.data_source_outputs')
     def update_combobox_values(self):
         available = self.variable_names_registry.data_source_outputs
         self._combobox_values = [''] + available
         self.name = ('' if self.name not in available else self.name)
 
+    def _get_label(self):
```
corranwebster: You might want to add a `@cached_property` decorator to the getter,

force_wfmanager/left_side_pane/tests/test_kpi_specification_model_view.py
```diff
@@ -44,3 +44,12 @@ def test_update_combobox_values(self):
 
     def test_label(self):
         self.assertEqual(self.kpi_specification_mv.label, "KPI")
+
+    def test_name_change(self):
+        self.data_source1.output_slot_info = [OutputSlotInfo(name='T1')]
+        self.kpi_specification_mv.model.name = ''
```
corranwebster: I'd add an:
```
with self.assertTraitChanges(self.kpi_specification_mv, 'label', count=1):
```
around this.  It does require adding `UnittestTools` as we talked about yesterday.

force_wfmanager/left_side_pane/tests/test_kpi_specification_model_view.py
```diff
@@ -44,3 +44,12 @@ def test_update_combobox_values(self):
...
+        self.assertEqual(self.kpi_specification_mv.label,
+                         'KPI')
+        self.kpi_specification_mv.model.name = 'T1'
```
corranwebster: Same as line 50.

force_wfmanager/left_side_pane/tests/test_kpi_specification_model_view.py
```diff
@@ -44,3 +44,12 @@ def test_update_combobox_values(self):
...
+        self.assertEqual(self.kpi_specification_mv.label,
+                         'KPI: '+self.kpi_specification_mv.model.name)
```
corranwebster: Don't compute the answer if you know precisely what it will be (I think "KPI: T1" in this case.)

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -225,12 +225,15 @@ def _fill_slot_rows(self, input_slots, output_slots):
         """ Fill the tables rows according to input_slots and output_slots
         needed by the evaluator and the model slot values """
         available_variables = self._get_available_variables()
-
         input_representations = []
+
         for index, input_slot in enumerate(input_slots):
+            new_name = self.model.input_slot_info[index].name
             slot_representation = InputSlotRow(model=self.model, index=index)
+            new_name = (new_name if new_name in available_variables else '')
```
corranwebster: Nit picky style thing: I'd just do this as
```
if new_name not in available_slots:
     new_name = ''
```
as I think it reads a little more clearly; either that or assign `slot_representation.name = (new_name if new_name in available_variables else '')` directly on L235 below.

corranwebster: Also, I would probably group the lines dealing with the `new_name` variable together to avoid context switching when reading the code.

corranwebster:
Oh, and one more thing, rather than adding the issue number in the title, if you put something like "fixes #160" in the initial description, github will link to the issue and will automatically close the issue when the PR is merged.

Issue #142, opened 2018-05-10
Factories in the modal dialog should be divided by provenance.


stefanoborini:
Right now all factories are shown together. It would be better to show factories partitioned according to their plugin of origin

PR #177, opened 2018-06-18
Factories grouped by creator using TreeEditor

jjenthought:
Addresses #142 
This creates separate lists of factories for each plugin. Initially, there is a list of all factories - regardless of their source plugin. Since each uitem binds to one traits object (afaik), this means creating a new traitlist for each plugin. If a uitem can be bound to a subset of a list, this would not be necessary.
Also requiring review is how we get the plugin names. At the minute it's done by module name as i think it's possible to have overlapping names if this is done via plugin.id. It may be better to go with plugin.id and change how this is defined so the value is unique.
This also needs tests to be added.


jjenthought:
That’s fine, if you’re free at some point today we can discuss better ways to do it. There were a few things I was uncertain about while writing it so it’ll be good to clear things up!
James

jjenthought:
@stefanoborini Github isn't showing me the notes that were added to the code, let me know if I've missed any changes we talked about. 
Unfortunately I can't find anything in tree_editor.py to prevent it automatically creating a VSplit as part of the editor itself.

PR #178, opened 2018-06-18
Open file from command line

### LINKED ISSUE #147: Possibility to open file from the command line
It would be nice to be able to specify the file to open at the command line, so we can start `force_wfmanager filename.json` instead of having to start the empty application, then open the file.

jjenthought:
This works fine at present, but it would be good to know if there's a better way to pass through arguments. At the minute it always passes a filename argument, which is either None or a filepath string. It's not too bad for just one input, but if there are going to be a lot of command line options it may need restructuring.
Fixes #147

### DIFF
--- a/force_wfmanager/gui/run.py
+++ b/force_wfmanager/gui/run.py
@@ -21,15 +21,14 @@ push_exception_handler(lambda *args: None, reraise_exceptions=True)
 
 @click.command()
 @click.version_option(version=__version__)
-@click.option('--file', '-f', 'workflow_file', default=None, type=click.Path(),
-              help='Loads a previously saved workflow '
-                   'from the given file path')
+@click.argument('workflow_file', type=click.Path(exists=True), required=False,
+                default=None,)
 def force_wfmanager(workflow_file):
     """Launches the FORCE workflow manager application"""
     main(workflow_file=workflow_file)
 
 
-def main(workflow_file=None):
+def main(workflow_file):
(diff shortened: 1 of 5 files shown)


force_wfmanager/gui/run.py
```diff
@@ -21,18 +21,20 @@
 
 @click.command()
 @click.version_option(version=__version__)
-def force_wfmanager():
+@click.option('--file', '-f', 'workflow_file', default=None, type=click.Path(),
```
stefanoborini: why an option? Use an argument.

force_wfmanager/wfmanager_plugin.py
```diff
@@ -27,4 +34,20 @@ def _create_task(self):
         factory_registry = self.application.get_plugin(
             FACTORY_REGISTRY_PLUGIN_ID)
 
-        return WfManagerTask(factory_registry=factory_registry)
+        wf_manager_task = WfManagerTask(factory_registry=factory_registry)
```
stefanoborini: You already have this code in the WfManagerTask. Why are you copying it here. The wfmanagertask is responsible for opening a given file, you just have to refactor the UI action in two parts: one that responds to the event, the other that accepts the filename and loads the file.

stefanoborini:
@jjenthought I didn't test it...

jjenthought:
Sorry I thought you meant that everything was ready to be merged

Issue #167, opened 2018-06-11
Disable run button until validation passes


stefanoborini:
The Run button should be disabled until the validation of completed and successful.

Issue #166, opened 2018-06-11
Add verification icons


stefanoborini:
We want to have verification icons that tell you if a given workflow has problems and where.

stefanoborini:
This is complicated by two factors: the first is that the verify will return you which model object is faulty, but you have to assign the flag to the modelview instead.

The second problem is that for every individual change to the model, you have to rerun the verification, which requires either a complicated triggering, or to rework the model to bubble up its modification notification to the workflow.

stefanoborini:
I suggest to go with the second method we talked yesterday: any action that modifies the model triggers a verification in the action object. It's easier to implement and you stay confined to wfmanager.

PR #194, opened 2018-06-27
Gui workflow verification 

jjenthought:
Introduces the workflow verification with local and global errors.

For issues #166, #167


force_wfmanager/left_side_pane/kpi_specification_model_view.py
```diff
@@ -34,13 +36,20 @@ class KPISpecificationModelView(ModelView):
...
+    verify_workflow_event = Event
+
     def __init__(self, model, variable_names_registry, **kwargs):
         super(KPISpecificationModelView, self).__init__(
             model=model,
             variable_names_registry=variable_names_registry,
             **kwargs
         )
 
+    @on_trait_change('model.name, model.objective')
```
stefanoborini: remove the space.

force_wfmanager/left_side_pane/mco_parameter_model_view.py
```diff
@@ -17,12 +17,22 @@ class MCOParameterModelView(ModelView):
...
     traits_view = View(
         Item("model.name"),
         Item("model.type"),
         kind="subpanel",
     )
 
+    #: Event to request a verification check on the workflow
+    verify_workflow_event = Event
+
+    @on_trait_change('model.name, model.type')
```
stefanoborini: space

force_wfmanager/left_side_pane/side_pane.py
```diff
@@ -60,6 +62,10 @@ def _workflow_tree_default(self):
             model=self.workflow_m
         )
 
+    @on_trait_change('workflow_tree.workflow_mv.valid')
+    def enable_run(self):
```
stefanoborini: update_run_btn_status

force_wfmanager/left_side_pane/workflow_model_view.py
```diff
@@ -33,6 +37,17 @@ class WorkflowModelView(ModelView):
     #: Defines if the Workflow is valid or not
     valid = Bool(True)
 
+    label = Str()
```
stefanoborini: what is this label for? needs comment

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -56,123 +59,139 @@
 edit_data_source_action = Action(name='Edit...', action='edit_data_source')
 
 
+#: Wrapper to call workflow verification after each method
+def verify(func):
```
stefanoborini: I'd call it verify_workflow()

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -145,6 +151,11 @@ def __init__(self, model, variable_names_registry, *args, **kwargs):
 
         self._create_slots_tables()
 
+    @on_trait_change('input_slots_representation.name,'
+                     'output_slots_representation.name')
+    def datasource_change(self):
```
stefanoborini: for consistency, use data_source

force_wfmanager/left_side_pane/notification_listener_model_view.py
```diff
@@ -12,6 +12,11 @@ class NotificationListenerModelView(ModelView):
     #: Label to be used in the TreeEditor
     label = Str()
 
+    error_message = Str()
+
+    #: Event to request a verification check on the workflow
+    verify_workflow_event = Event
+
```
stefanoborini: I suspect you need a handler here too to check if something changes and trigger the event

force_wfmanager/left_side_pane/side_pane.py
```diff
@@ -43,13 +43,15 @@ class SidePane(TraitsDockPane):
     #: Used when the computation is running
     enabled = Bool(True)
```
stefanoborini: We will have to come back and fix this to "ui_enabled" to clarify. Not now, maybe open an issue for it.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -283,16 +334,154 @@ def new_data_source(self, ui_info, object):
...
+        else:
+            self.selected_error = self.selected_mv.error_message
+
+    @on_trait_change("workflow_mv.verify_workflow_event")
+    def received_verify_request(self):
+        """Checks if the root node of workflow tree is requesting a
+        verification of the workflow"""
+        self.verify_workflow_event = True
+
+    def _verify_workflow_event_fired(self):
```
stefanoborini: be wary of the _fired or _changed methods, they are known to bite because they bind at a different time. Stay consistent and always use on_trait_change.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -283,16 +334,154 @@ def new_data_source(self, ui_info, object):
...
+                        ExecutionLayerModelView: ['data_sources_mv']}
+        verify_tree(self.workflow_mv, parent_child, result)
+
+
+def verify_tree(mv_list, mappings, errors):
+    """ Call with a list of modelview objects to assign
+    their error messages, plus the error message of
+    their subsequent child modelviews.
+
+    :param mv_list
```
stefanoborini: We use a different style for docstrings. sphinx is not configured to pick these

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -283,16 +334,154 @@ def new_data_source(self, ui_info, object):
...
+    :param errors
+    A list of VerifierError, where object.error is an error message and
+    object.subject is the component of the workflow it applies to
+    """
+    if not isinstance(mv_list, list):
+        mv_list = [mv_list]
+
+    for modelview in mv_list:
+        modelview.error_message = ''
+        for type, lists in mappings.items():
```
stefanoborini: type is a builtin, so it's good practice to use type_ instead.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -283,16 +334,154 @@ def new_data_source(self, ui_info, object):
...
+                    child_mv_list = getattr(modelview, child_mv_list_name)
+                    child_errors = verify_tree(child_mv_list,
+                                               mappings,
+                                               errors)
+                    modelview.error_message += child_errors
+                    if modelview.error_message != '':
+                        modelview.valid = False
+
+    errors_from_list = ''
+
```
stefanoborini: From here on, you lose me. We need to talk about the rest because I can't understand it.

force_wfmanager/left_side_pane/notification_listener_model_view.py
jjenthought: I'm not sure what to put as the check here at the minute. The base notification listeners only have one trait, which is their factory.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -56,123 +59,139 @@
...
+    return wrap
+
+
 class TreeNodeWithStatus(TreeNode):
     """ Custom TreeNode class for workflow elements """
+
     def get_icon(self, object, is_expanded):
         return 'icons/valid.png' if object.valid else 'icons/invalid.png'
 
+    original_label_changed = TreeNode.when_label_changed
```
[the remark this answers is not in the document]
jjenthought: Actually I was meaning to ask you about this! I basically wanted to 'decorate' the when_label_changed method in TreeNode to also be triggered when the TreeNode changes from valid to invalid. But I don't know if that's possible. 
Instead I'm having to override the original when_label_changed from TreeNode, and before I do that I have to copy the original function to a local variable so I don't lose the original functionality where it detects changes to a label in the TreeNode

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -200,8 +236,12 @@ def _workflow_mv_default(self):
 
     @on_trait_change('model')
```
jjenthought: If there is a way to decorate an on_trait_change with @verify_wkflow, it would be nice as at the minute verify_workflow_event has to be triggered within the function

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -283,16 +334,178 @@ def new_data_source(self, ui_info, object):
...
+            # Combine errors which refer to similar problems
+            current_modelview.error_message = collate_errors(current_modelview.
+                                                             error_message)
+        else:
+            current_modelview.valid = True
+
+        return current_modelview.error_message
+
+
+def collate_errors(error_message):
```
jjenthought: This function is still a bit over-specific, but it works okay for now

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -122,6 +122,12 @@ class DataSourceModelView(ModelView):
     #: Defines if the evaluator is valid or not
     valid = Bool(True)
 
+    #: An error message for issues in this modelview
+    error_message = Str
```
stefanoborini: While technically there's no difference in using Str() vs Str, in practice I favour consistency of using () every time, as it's mandatory when you pass information. Also, traitlets (similar to traits) always uses (), so it's good practice that you don't get too attached to that pattern (we do have some code that uses traitlets)

force_wfmanager/left_side_pane/notification_listener_model_view.py
stefanoborini: True. However we will potentially have to deal with validity of the model itself as well. Leave it as is for now, but for example if the model does require a password and it's not there, we will have to mark it as invalid. This makes it even more complex because validity is decided by the model itself...

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -69,119 +71,25 @@
                                  enabled_when=call_modelview_editable)
 
 
+#: Wrapper to call workflow verification after each method
+def verify_wkflow(func):
```
stefanoborini: workflow

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +367,203 @@ def new_data_source(self, ui_info, object):
...
         self.workflow_mv.remove_execution_layer(object.model)
 
+    @on_trait_change("workflow_mv.verify_workflow_event")
+    def received_verify_request(self):
+        """Checks if the root node of workflow tree is requesting a
+        verification of the workflow"""
+        self.verify_workflow_event = True
+
+    @on_trait_change("verify_workflow_event")
+    def perform_verify_workflow_event_(self):
```
stefanoborini: extra _ at the end

force_wfmanager/left_side_pane/workflow_tree.py
jjenthought: I ended up changing from @verify_workflow to this because there's a function I use which is  also called verify_workflow. Is there any particular convention to go by for this sort of thing?

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +367,203 @@ def new_data_source(self, ui_info, object):
...
+
     def modelview_editable(self, modelview):
         """Checks if the model associated to a ModelView instance
         has a non-empty, editable view """
         return model_info(modelview.model) != []
+
+    @on_trait_change("selected_mv,selected_mv.error_message,selected_mv.label")
+    def update_selected_error(self):
+
+        mv_label = self.selected_mv.label
```
stefanoborini: if self.selected_mv is None you can't extract the label.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +367,203 @@ def new_data_source(self, ui_info, object):
...
+                "Errors for {}:".format(mv_label), body_strings)
+
+
+def collate_errors(error_message):
+    """Group together similar error messages. For example, if output
+    parameters 1,2 and 3 have undefined names display 'Output parameters 1-3
+    have undefined names', rather than 3 separate error messages."""
+
+    #: List of error messages
+    error_list = error_message.split('\n')
```
stefanoborini: You should pass the list of error messages instead of splitting the already formed string. Compose the final string only at the very end, before displaying.

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +367,203 @@ def new_data_source(self, ui_info, object):
...
+        else:
+            error_list = self.selected_mv.error_message.split('\n')
+            body_strings = ''.join([HTML_SINGLE_ERROR.format(error)
+                                    for error in error_list])
+            self.selected_error = HTML_ERROR_TEMPLATE.format(
+                "Errors for {}:".format(mv_label), body_strings)
+
+
+def collate_errors(error_message):
+    """Group together similar error messages. For example, if output
```
stefanoborini: this is clever, but dangerous. You are reconstructing high level knowledge from degraded information (the error string). The likelihood that it will survive complex future input is low. Keep it because it seems to work and it's actually very cleverly implemented, but be wary of that in the future.

I suggest adding tests for it. If we want to support this strategy, we should instead provide high level support for it at the level of verify_workflow

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +367,203 @@ def new_data_source(self, ui_info, object):
...
+                                              'notification_listeners_mv',
+                                              'execution_layers_mv'],
+                        'MCOModelView': ['mco_parameters_mv', 'kpis_mv'],
+                        'ExecutionLayerModelView': ['data_sources_mv']}
+
+        # Communicate the verification errors to each level of the
+        # workflow tree
+        self.verify_tree(parent_child, errors)
+
+    def verify_tree(self, mappings, errors, current_modelview=None):
```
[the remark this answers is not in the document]
jjenthought: Would adding metadata to the modelview lists be a good way to achieve this in a more automatic way? I found checking their trait types directly was a bit of a mess, though the information is clearly there somewhere for validation etc.

force_wfmanager/left_side_pane/workflow_tree.py
jjenthought: Yeah I originally tried to do this more generally but couldn't think of a good way. So it relies on the first number in an error message being indicative of some common property, which like you say may not be true in future!

jjenthought:
This now requires https://github.com/force-h2020/force-bdss/pull/171 to work.

force_wfmanager/left_side_pane/workflow_tree.py
stefanoborini: ah damn... true.

triggers_verify then. 

The reason is that if you stay consistent with naming it's easier to grep around. Once you pick a name (e.g. workflow) stay consistent with it as much as possible, unless it's very encapsulated (e.g. a local variable called wf to store a workflow is fine)

force_wfmanager/left_side_pane/mco_parameter_model_view.py
```diff
@@ -12,17 +12,39 @@ class MCOParameterModelView(ModelView):
...
+    #: Event to request a verification check on the workflow
+    verify_workflow_event = Event
+
+    @on_trait_change('model.name,model.type')
+    def parameter_change(self):
+        self.verify_workflow_event = True
+
+    def _get_label(self):
+        if self.model.name != '':
+            label_name = 'Name: '+self.model.name
```
stefanoborini: change the representation to match the kpi, like

Range: TYPE name
KPI: name (TARGET)

force_wfmanager/left_side_pane/side_pane.py
```diff
@@ -59,6 +61,10 @@ def _workflow_tree_default(self):
             model=self.workflow_m
         )
 
+    @on_trait_change('workflow_tree.workflow_mv.valid')
+    def update_run_btn_status(self):
```
stefanoborini: I don't know why but the run button is still enabled..

force_wfmanager/left_side_pane/side_pane.py
```diff
@@ -59,6 +61,10 @@ def _workflow_tree_default(self):
             model=self.workflow_m
         )
 
+    @on_trait_change('workflow_tree.workflow_mv.valid')
```
stefanoborini: you can bind directly to the workflow_tree, as the event is propagated anyway

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -317,21 +364,143 @@ def new_data_source(self, ui_info, object):
...
     def modelview_editable(self, modelview):
         """Checks if the model associated to a ModelView instance
         has a non-empty, editable view """
         return model_info(modelview.model) != []
+
+    @on_trait_change("selected_mv,selected_mv.error_message,selected_mv.label")
+    def update_selected_error(self):
+
+        # If nothing is currently selected, display all the errors
+        if self.selected_mv is None:
```
stefanoborini: you recurse here.

Issue #169, opened 2018-06-11
Show errors in form of dialogs, in addition to logging


stefanoborini:
When an error occurs in the plugins, we should show a dialog that reports that there was an error.

PR #199, opened 2018-07-05
Using Basefactory in NewModalDialog

### LINKED ISSUE #198: Use BaseFactory Instances
Where there are expressions using Either(Instance(Factory1),Instance(Factory2)) in the code, use Instance(BaseFactory) instead.

jjenthought:
Relies on https://github.com/force-h2020/force-bdss/pull/169, fixes #198


stefanoborini:
It is unfortunate that we still need to import the specialised ones to find the plugin. We'll have to provide an appropriate method so that plugin can be found transparently, but not now.

stefanoborini:
@jjenthought actually, I checked and we do have the plugin available also in the parameter factories, so you can technically get rid of the part where it differentiates in order to find the plugin. If you don't understand which one, come talk to me.

PR #200, opened 2018-07-09
Randomise zmq port

### LINKED ISSUE #164: Randomise the ZMQ port to reduce conflict with antivirus
Apparently sophos likes to open a lot of ports for whatever reason, and the result is that the likelihood that the port is taken when the ZMQ server starts is high, resulting in the following error

```
  self._data = newdata
Exception in thread ZMQServer:
  File "/Users/sborini/.edm/envs/force/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/Users/sborini/Work/github/force-h2020/force-wfmanager/force_wfmanager/server/zmq_server.py", line 69, in run
    self._inproc_socket) = self._setup_sockets()
  File "/Users/sborini/Work/github/force-h2020/force-wfmanager/force_wfmanager/server/zmq_server.py", line 138, in _setup_sockets
    pub_socket.bind(self.config.pub_url)
  File "zmq/backend/cython/socket.pyx", line 495, in zmq.backend.cython.socket.Socket.bind (zmq/backend/cython/socket.c:5324)
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:7916)
zmq.error.ZMQError: Address already in use
```

We need to randomise the port and check if it's occupied before selecting it. It should not create any problem as the information about the server is passed to the executor in the file.

stefanoborini:
Randomise the zeroMQ port so that less chance of conflict with the antivirus or other services open at the same port are reduced.

Fixes #164


force_wfmanager/server/zmq_server.py
```diff
@@ -127,23 +201,47 @@ def stop(self):
...
         sync_socket.setsockopt(zmq.LINGER, 0)
-        sync_socket.bind(self.config.sync_url)
+        sync_port = sync_socket.bind_to_random_port("tcp://*")
 
         inproc_socket = context.socket(zmq.PAIR)
         inproc_socket.bind("inproc://stop")
-        return pub_socket, sync_socket, inproc_socket
+        return pub_socket, pub_port, sync_socket, sync_port, inproc_socket
+
+    def _close_network_sockets_noerror(self):
```
jjenthought: Just a naming thing, close_network_sockets_noerror is used for both closing sockets normally and after an error. 
I suppose now there's a distinct on_error_callback we could have a method for each case, but it might be easier just to rename it close_network_sockets and keep things as they are at the minute.

stefanoborini: it is mostly referred to the fact that the method cannot raise. The close operation is quite delicate, as it must be done in any case, and can fail if there was a failure. I will call it noexc instead.

Issue #161, opened 2018-06-08
Limit the variables to the appropriate types


stefanoborini:
Combobox selection should only give the variables matching the type (e.g. a VOLUME slot should display only VOLUME variables). This will have to change  to a more refined mechanism when we introduce the EMMC types, but for now a simple string comparison is what we need.

jjenthought:
Closed by #201

PR #201, opened 2018-07-10
Variable type consistency

jjenthought:
Addresses #161. Note: The way the code extracts types at the minute is not correct.


force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -260,12 +260,21 @@ def update_slot_info_names(self):
...
+                     'input_slots_representation.type,')
     def update_data_source_input_rows(self):
-        available_variables = self._get_available_variables()
         for input_slot_row in self.input_slots_representation:
+            available_variables = self._get_available_variables(
+                input_slot_row.type)
             input_slot_row.available_variables = available_variables
 
-    def _get_available_variables(self):
+    def _get_available_variables(self, var_type=None):
```
jjenthought: If the type of a input variable is empty the set of available variables does not restrict based on type. Though this should not be something which occurs the way we have things set up now, as datasource input types are not settable afaik.

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -260,12 +260,21 @@ def update_slot_info_names(self):
                              self.output_slots_representation):
             row.name = info.name
 
-    @on_trait_change('variable_names_registry.available_variables[]')
+    @on_trait_change('variable_names_registry.available_variables[],'
+                     'output_slots_representation.name,'
+                     'output_slots_representation.type,'
+                     'input_slots_representation.name,'
+                     'input_slots_representation.type,')
```
stefanoborini: Careful with this end comma. It might not lead to any problems but it's hard to say.

force_wfmanager/left_side_pane/data_source_model_view.py
stefanoborini: types are going to be cuba types. they will have their own verification for compatibility, so I'd take it out.

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -19,22 +20,28 @@ class VariableNamesRegistry(HasStrictTraits):
...
+    #: - the first layer has available "pikachu" and "squirtle".
+    #: - the second layer has available "pikachu" and "squirtle",
+    #: because the first layer added nothing (hence the empty second list)
+    #: - the third layer has available "pikachu", "squirtle" and "scyther".
     #:
     #: The size of the base list should be the number of layers plus one.
     #: the last one being the variables that are added by the last layer.
-    available_variables_stack = List(List(Identifier))
+
+    available_variables_stack = List(List(Dict(Identifier, CUBAType)))
```
stefanoborini: this is getting too big to parse. Try to extract a trait like LayerVariables. It might help with understanding.

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -50,50 +57,68 @@ def __init__(self, workflow, *args, **kwargs):
 
     @on_trait_change(
         'workflow.mco.parameters.name,'
-        'workflow.execution_layers.data_sources.output_slot_info.name')
+        'workflow.execution_layers.data_sources.output_slot_info.name,'
+        'workflow.mco.parameters.type,'
```
stefanoborini: comma

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -50,50 +57,68 @@ def __init__(self, workflow, *args, **kwargs):
...
         # At the first layer, the available variables are the MCO parameters
         if self.workflow.mco is None:
             stack.append([])
         else:
-            stack.append([
-                p.name
-                for p in self.workflow.mco.parameters
-                if len(p.name) != 0
-            ])
+            stack.append([{'name': p.name, 'type': p.type}
```
stefanoborini: I am not sure about this. I assumed you had a mapping from identifier to type, that is p.name: p.type.

force_wfmanager/left_side_pane/variable_names_registry.py
```diff
@@ -50,50 +57,68 @@ def __init__(self, workflow, *args, **kwargs):
...
 
         for layer in self.workflow.execution_layers:
             stack_entry_for_layer = []
             for ds in layer.data_sources:
-                stack_entry_for_layer.extend([
-                    info.name
-                    for info in ds.output_slot_info
-                    if info.name != ''])
+                ds_names = [info.name for info in ds.output_slot_info]
+                ds_output_slots = ds.factory.create_data_source().slots(ds)[1]
```
stefanoborini: this code is in the hands of the plugin implementer, so it could throw exceptions. uhm....

stefanoborini:
I'll have to take another look at it tomorrow

force_wfmanager/left_side_pane/variable_names_registry.py
jjenthought: I think it may not be as easy to have a VariableNames object if we want types as well. It works at the minute because we can get all the information from the Workflow, but we need data which can only be found in the ModelView(s). 
So I think the VariableNames functionality will have to be implemented across the ModelViews themselves as well as in VariableNames, due to the fact that mvs are recreated often. 
Also I realised that as well as relying on plugin code, calling get/create_data_source() does not account for slot changes, so it's not a good solution unfortunately!

PR #203, opened 2018-07-11
Do verbose tests

stefanoborini:
Run verbose tests so that we can get more information about potential noisy tests.


stefanoborini:
Self merging as it's trivial.

Issue #148, opened 2018-06-05
Ability to provide description to data sources input and outputs


stefanoborini:
Right now the table with the input and outputs of the datasources just reports the type and an empty space for the variable name to do the binding. Without any additional information, it's impossible to tell what each value means.

We might have to rework the design of the GUI to accommodate for a longer description.

stefanoborini:
Related to #141

PR #204, opened 2018-07-11
Add Descriptions to datasource rows

jjenthought:
Addresses #148 
This works & can go in for now but will need changes in terms of how everything is visually Since we're planning to make changes to the UI in the near future it might be sensible to leave it till that's decided.


force_wfmanager/left_side_pane/view_utils.py
```diff
@@ -44,3 +45,13 @@ def _item_info_from_group(group_contents, item_info=None):
         elif isinstance(entity, Item):
             item_info.append(entity.name)
     return item_info
+
+
+def get_default_background_color():
+    """Return the default background color for a Qt Application"""
+    palette = QtGui.QPalette()
+    rgba_color = palette.window().color().getRgb()
```
stefanoborini: This might work, but I would have used the QApplication.palette().getColor(Active, Background)

They _might_ be equivalent, because the new palette uses the application palette by default as a prototype, but I am not 100% sure about the window() call.

my Qt is rusty though.

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -82,6 +84,7 @@ def update_name(self):
 output_slots_editor = TableEditor(
     sortable=False,
     configurable=False,
+    selected="selected_slot_row",
```
stefanoborini: it felt a bit weird to have the same selected entry for two tables, but apparently table editor is smart enough to unselect the other one when it's not in the list.

force_wfmanager/left_side_pane/view_utils.py
stefanoborini: Question: what happens if you don't specify a background color? will it default to white? We might not need this.

stefanoborini:
Ok for me, but you might have a point on the Qt thing I am not familiar with.

stefanoborini:
Ok, I am curious about something. Why do you extract and store the desc independently? Could you just access it through the selected row?

force_wfmanager/left_side_pane/data_source_model_view.py
```diff
@@ -280,3 +309,83 @@ def update_data_source_input_rows(self):
...
+            .right-col{{
+                width: 15%;
+                display: inline-block;
+                word-wrap: break-word;
+            }}
+        </style>
+    </head>
+    <body>
+        <div class="container">
+            <div class="left-col" w>
```
stefanoborini: extra w

PR #206, opened 2018-07-12
Plugin window showing installed plugins

### LINKED ISSUE #170: Have a window that shows which plugins are installed, which ones failed, and why.
In the menubar, we should have an entry that opens a window. This window will list all the available plugins, and if there was an error in loading it, it will show an icon and information about the nature of the problem.

stefanoborini:
Fixes #170


stefanoborini:
@jjenthought this is ready for review.

force_wfmanager/plugin_dialog.py
```diff
@@ -0,0 +1,119 @@
...
+
+class PluginDialog(HasStrictTraits):
+    """Displays a list of the available plugins, together with their
+    description.
+    """
+    plugins = List(Instance(BaseExtensionPlugin))
+
+    selected_plugin = Instance(BaseExtensionPlugin)
+
+    selected_plugin_HTML = Unicode()
```
jjenthought: I was going to suggest this should be HTML(), but looking into it HTML has the parent class of String (rather than BaseUnicode).

PR #207, opened 2018-07-13
Python terminal as docked entry

stefanoborini:
I cannot find any reason why it should not work, yet it does not show anything.
@corranwebster  might be of help?


corranwebster:
Tasks saves layout information in your app's configuration directory (often `.enthought/appname` but it could be elsewhere depending on `ETSConfig`) its common that new panes don't show because they aren't in your saved task layout.

stefanoborini:
@corranwebster I tried that without any luck. Ioannis and I spent some time figuring it out with no result. I decided for a workaround which seems to work, having an additional tab with the ShellEditor instead. However, I cannot find a way to obtain the application from the task.

corranwebster:
Not sure what the problem is - another possible work around would have been to add the [DockPaneToggleGroup](https://github.com/enthought/pyface/blob/master/pyface/tasks/action/dock_pane_toggle_group.py) to your menus and try explicitly toggling the visibility.

Issue #186, opened 2018-06-20
Plot is not reset when calculation is reissued


stefanoborini:
If you perform an evaluation, then perform it again, the table is reset, but the plot preserves the old points until the first new point arrives.

jjenthought:
Closed by #210

Issue #185, opened 2018-06-20
Perform autoscaling of plot when changing axes


stefanoborini:
The plot should autoscale to the appropriate scales when changing the X and Y values.

jjenthought:
Closed by #210

Issue #184, opened 2018-06-20
Plot should show the appropriate x and y axis names.


stefanoborini:
The plot currently does not show anything in the x and y axes names. We need to add this information.
One problem is that we don't have anything related to units, so we can't include that information.

jjenthought:
closed by #210

PR #210, opened 2018-07-17
Plot Display Resizing

jjenthought:
Addresses #184, #185 & #186. While the plot did seem to get resized automatically, it sets the plot boundary to be exactly on an outlying plot point. This adds a little space around these edge values (10% of the plot height/width) and provides a button to reset the plot view. 
This also resets the plot view when there are no data value names available (and therefore no data to plot), for example after the analysis model has been cleared.

### DIFF
--- a/force_wfmanager/central_pane/plot.py
+++ b/force_wfmanager/central_pane/plot.py
@@ -1,5 +1,5 @@
 from traits.api import (HasStrictTraits, List, Instance, Enum, Tuple,
-                        on_trait_change)
+                        on_trait_change, Button, Bool, Property)
 
 from traitsui.api import View, UItem, Item, VGroup, HGroup
 
@@ -36,12 +36,18 @@ class Plot(HasStrictTraits):
     #: Datasource of the plot (used for selection handling)
     _plot_index_datasource = Instance(ArrayDataSource)
 
+    #: Button to reset plot view
+    reset_plot = Button('Reset View')
+
+    reset_enabled = Property(Bool(False), depends_on="_plot_data")
+
     view = View(VGroup(
         HGroup(
             Item('x'),
             Item('y'),
         ),
         UItem('_plot', editor=ComponentEditor()),
+        UItem('reset_plot', enabled_when='reset_enabled')
     ))
 
     def __init__(self, analysis_model, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


force_wfmanager/central_pane/plot.py
```diff
@@ -174,8 +182,46 @@ def _update_plot_data(self):
...
+            y_size = abs(y_max - y_min)
+            y_max = y_max + 0.1 * abs(y_size)
+            y_min = y_min - 0.1 * abs(y_size)
+
+            self._plot.range2d.x_range.low = x_min
+            self._plot.range2d.x_range.high = x_max
+            self._plot.range2d.y_range.low = y_min
+            self._plot.range2d.y_range.high = y_max
+
+            return x_min, x_max, y_min, y_max
```
jjenthought: This return is pretty much just for information as the actual assignment has already taken place. I don't know if there are any conventions we follow about not mixing return lengths or anything like that

force_wfmanager/central_pane/plot.py
```diff
@@ -36,12 +36,20 @@ class Plot(HasStrictTraits):
     #: Datasource of the plot (used for selection handling)
     _plot_index_datasource = Instance(ArrayDataSource)
 
+    #: Button to reset plot view
+    reset_plot = Button('Reset View')
+
+    reset_enabled = Property(Bool(False), depends_on="_plot_data")
```
stefanoborini: Passing False here does not help. A property is always invoked to determine its value.

force_wfmanager/central_pane/plot.py
```diff
@@ -36,12 +36,20 @@ class Plot(HasStrictTraits):
...
+
     view = View(VGroup(
         HGroup(
             Item('x'),
             Item('y'),
         ),
         UItem('_plot', editor=ComponentEditor()),
+        VGroup(
+            UItem('reset_plot', enabled_when='reset_enabled')
```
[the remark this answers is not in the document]
stefanoborini: What I would expect (but not sure) is that enabled_when is invoked when the view is shown. That triggers your property get method, which in turns forces the _plot_data_default to be invoked. I don't see any problem with that. Did you get a specific exception?

stefanoborini:
There's something I don't understand. It seems that if one zooms out after resetting, there is a max zoom out value set. Might be a bug of plot though. We'll check it together tomorrow.

force_wfmanager/central_pane/plot.py
```diff
@@ -195,3 +242,20 @@ def update_model(self):
...
+
+    def _get_reset_enabled(self):
+        x_data = self._plot_data.get_data('x')
+        if len(x_data) > 1:
+            return True
+        return False
+
+    def set_plot_range(self, x_low, x_high, y_low, y_high):
+        """Helper method to set the size of the current _plot"""
+        self._plot.range2d.x_range.low = x_low
```
stefanoborini: I suspect my comment is because of this. this clamps the zoom for some reason.

jjenthought: So if you zoom out to the maximum and click reset, then try and zoom out within the newly reset plot it ... doesn't do anything. 
Perhaps there's some internal counter in chaco for how many 'zoom out' actions you're allowed which is not reset by resizing the plot?

force_wfmanager/central_pane/plot.py
```diff
@@ -166,7 +180,7 @@ def _update_plot_data(self):
         """Set the plot data model to the appropriate arrays so that they
         can be displayed when either X or Y selections have been changed.
         """
-        if self.x is None or self.y is None:
+        if self.x is None or self.y is None or self._data_arrays == []:
```
jjenthought: Fixes the issue when update_data_arrays gets called with value_names set, but no available data ready

force_wfmanager/central_pane/plot.py
```diff
@@ -177,6 +191,46 @@ def _update_plot_data(self):
         self._plot_data.set_data('x', self._data_arrays[x_index])
         self._plot_data.set_data('y', self._data_arrays[y_index])
 
+        self.resize_plot()
+
+    def resize_plot(self):
```
jjenthought: This could porbably be done using chaco settings, but the automatic versions didn't seem to work (passing tight_bounds=False). There was also an option to provide your own function which takes (x-low, x-high, margin, tight_bounds) and returns the plot range (min, max).
But then it would basically be this function, except you couldn't call it in the case where you wanted to manually reset the plot view.

PR #213, opened 2018-07-17
Hide ui notification

jjenthought:
Addresses https://github.com/force-h2020/force-wfmanager/issues/187. The change in the pr makes it so we now only create a modelview for models whose factory has ui_visible set to True. This is the default case in base_notification_listener_factory.


stefanoborini:
Good solution for now, but I am considering we should make a copy for file dumping anyway. Not now though. Good for me

PR #222, opened 2018-07-23
Refactor new_modal_dialog

jjenthought:
A few small changes: Added Unicode, consistently used " over ' for strings, more sensible variable names and HTML strings are now created using a function rather than using .format within a function.

### DIFF
--- a/force_wfmanager/left_side_pane/new_entity_modal.py
+++ b/force_wfmanager/left_side_pane/new_entity_modal.py
@@ -1,6 +1,7 @@
 
+from __future__ import unicode_literals
 from traits.api import (HasStrictTraits, Instance, List, Either,
-                        on_trait_change, Dict, Str, Property, HTML, Bool,
+                        on_trait_change, Dict, Property, Unicode, Bool,
                         ReadOnly)
 from traitsui.api import (View, Handler, HSplit, Group, VGroup, UItem,
                           InstanceEditor, OKCancelButtons, Menu,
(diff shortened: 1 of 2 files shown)


force_wfmanager/left_side_pane/new_entity_modal.py
```diff
@@ -91,37 +83,37 @@ class NewEntityModal(HasStrictTraits):
...
     OKCancelButtons[0].trait_set(enabled_when="selected_factory is not None")
 
-    model_description_HTML = Property(HTML, depends_on="current_model")
+    model_description_HTML = Property(Unicode, depends_on="current_model")
 
-    current_model_editable = Property(Bool, depends_on='current_model')
+    current_model_editable = Property(Bool, depends_on="current_model")
 
-    no_config_options_msg = ReadOnly(HTML)
+    no_config_options_msg = ReadOnly(Unicode)
```
stefanoborini: uhm... I am not sure, but I think that if we use HTML, we can skip using HTMLEditor() in the view. Do you have more information from Corran about possible different behavior? I used Unicode, but I might be wrong.

stefanoborini:
Try to improve coverage, there are some interesting spots.

force_wfmanager/left_side_pane/new_entity_modal.py
jjenthought: I haven't asked Corran, but in trait_types.py it's a subclass of String, with some metadata saying to use HTMLEditor. 
I was thinking it would be useful to have it as unicode, but I'll check whether this is a good idea!

PR #228, opened 2018-07-24
Close Exceptions MkII

jjenthought:
The previous fix #226 was causing the tasks to be deleted in _application_exiting_fired, before saving the application state. This happens in tasks_application.py:

def _prepare_exit(self):
     """ Called immediately before the extant windows are destroyed and the GUI event loop is terminated."""
    self.application_exiting = self
    self._save_state()

The fix is just to override _prepare_exit in wfmanager.py and switch the order of these two calls. The application_exiting event does not appear to be used anywhere else (searching the files in site-packages for "application_exiting" gives no results)

### DIFF
--- a/force_wfmanager/wfmanager.py
+++ b/force_wfmanager/wfmanager.py
@@ -13,12 +13,14 @@ class WfManager(TasksApplication):
             size=(800, 600)
         )]
 
+    # FIXME: This isn't needed if the bug in traitsui/qt4/ui_panel.py is fixed
     def _prepare_exit(self):
         """Same functionality as TasksApplication._prepare_exit(), but
         _save_state is called before application_exiting is fired"""
         self._save_state()
         self.application_exiting = self
 
+    # FIXME: This isn't needed if the bug in traitsui/qt4/ui_panel.py is fixed
     def _application_exiting_fired(self):
         for window in self.windows:
             tasks = window.tasks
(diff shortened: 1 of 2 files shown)


jjenthought:
I guess we should be careful with this one after the issues last time!

force_wfmanager/wfmanager.py
```diff
@@ -12,3 +12,17 @@ def _default_layout_default(self):
             active_task='force_wfmanager.wfmanager_task',
             size=(800, 600)
         )]
+
+    # FIXME: This isn't needed if the bug in traitsui/qt4/ui_panel.py is fixed
+    def _prepare_exit(self):
+        """Same functionality as TasksApplication._prepare_exit(), but
+        _save_state is called before application_exiting is fired"""
+        self._save_state()
+        self.application_exiting = self
```
stefanoborini: Could we argue that the current behavior is incorrect?

jjenthought:
Yeah I’d say we could. Application exiting is an event which gets fired but doesn’t seem to be handled by anything in envisage/traits etc. Which suggests it’s meant to be there for people writing stuff on top of the TasksApplication framework. I don’t know if it was expected that people wouldn’t directly access ‘internal’ stuff but if thats the case why have the event at all!
Unless of course I’ve missed somewhere it does get handled, but you’d think that would show up with grep.

corranwebster:
The `application_exiting` event is meant to be something that code (most likely plugins) can listen to so that they can handle any clean-up that requires the application to be fully running.

PR #230, opened 2018-07-25
Debug Logging Option

jjenthought:
Adds a cmd line option --debug which adds debug logs to force_wfmanager.log.

The only test is a run with debug=True passed to main. I'm not even sure if it should have this, since the logger is a system-wide thing, so any subsequent tests will also have debug logs written to force_wfmanager.log.

This doesn't cause any problems at the minute, but does mean the tests aren't totally self-contained.

### DIFF
--- a/force_wfmanager/gui/tests/test_click_run.py
+++ b/force_wfmanager/gui/tests/test_click_run.py
@@ -44,3 +44,5 @@ class TestClickRun(unittest.TestCase):
         with mock.patch('force_wfmanager.gui.run.WfManager') as mock_wf:
             mock_wf.return_value = MockWfManager()
             force_wfmanager.gui.run.main(debug=True)
+            self.log = force_wfmanager.gui.run.logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


force_wfmanager/gui/tests/test_click_run.py
```diff
@@ -30,3 +39,8 @@ def test_click_cli_main(self):
             force_wfmanager.gui.run.force_wfmanager()
 
             mock_run.force_wfmanager.assert_called()
+
+    def test_run_with_debug(self):
+        with mock.patch('force_wfmanager.gui.run.WfManager') as mock_wf:
+            mock_wf.return_value = MockWfManager()
+            force_wfmanager.gui.run.main(debug=True)
```
stefanoborini: You should check if the logging configuration level is actually set to debug.

jjenthought:
Ok, I’ve added a check - let me know if there’s a better way to access the log object. Getting it from the imported logging module seems a bit dodgy..

PR #234, opened 2018-08-01
Exit method moved to TaskWindow subclass

jjenthought:
This creates a subclass of TaskWindow which has an exit prompt, which gives the user a chance to save before quitting. 
Putting this functionality in the TaskWindow's close() method means it will also be triggered on closing via title bar, not just those performed through the menu bar. 
It currently relies on the first task in window.tasks being the WfManagerTask, as well as the window being the only TaskWindow (so closing it means exiting the application).

### DIFF
--- a/force_wfmanager/tests/test_wfmanager.py
+++ b/force_wfmanager/tests/test_wfmanager.py
@@ -1,12 +1,15 @@
 import unittest
-try:
-    import mock
-except ImportError:
-    from unittest import mock
+import unittest.mock as mock
 
+from envisage.core_plugin import CorePlugin
+from envisage.ui.tasks.tasks_plugin import TasksPlugin
+
+from pyface.api import (ConfirmationDialog, YES, NO, CANCEL)
+
+from force_wfmanager.wfmanager_plugin import WfManagerPlugin
 from force_wfmanager.wfmanager import WfManager, TaskWindowClosePrompt
 from force_wfmanager.wfmanager_task import WfManagerTask
-from pyface.api import (ConfirmationDialog, YES, NO, CANCEL)
+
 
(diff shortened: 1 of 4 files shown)


force_wfmanager/tests/test_wfmanager.py
```diff
@@ -1,9 +1,74 @@
 import unittest
+try:
+    import mock
+except ImportError:
+    from unittest import mock
```
stefanoborini: we need just this line. The try except was there for py2 compat

force_wfmanager/wfmanager_task.py
```diff
@@ -393,30 +392,10 @@ def _bdss_done(self, exception):
...
-            if not save_result:
-                return
-        elif result is CANCEL:
-            return
+        """ Exit the main window via TaskWindowClosePrompt's window.close(),
+        which opens a dialog with the option to save the current workflow."""
 
-        app = self.window.application
-        app.exit()
+        self.window.close()
```
stefanoborini: I am missing the connection between the TaskWindowClosePrompt and this class. What is the magic that makes it appear when you close?

jjenthought:
The TaskWindowClosePrompt is just a regular TaskWindow, but the close() method has the save dialog inserted before the usual method. Close is also vetoable and returns whether the window was destroyed or not, so it’s not changing the expected behaviour to have a close method which may not actually close the window! 

In wfmanager.py when you create the default TasksApplication, it (eventually) creates a window using the window_factory attribute. This is initially set to be a standard TaskWindow, so _window_factory_default is overridden to return TaskWindowClosePrompt.

This means that the window in the code has this different closing behaviour, and a title bar close will also trigger this method. The title bar close doesn't call app.exit(), so anything extra we put there will only get used for menu-bar closes (which could be useful if we wanted different behaviour at some point in future)

PR #239, opened 2018-08-28
Split Workflow - Tasks Level

jjenthought:
The main changes for this PR are that:
- We now have a results and a setup task. Most of the original functionality is in the setup task, at the minute the results one just mirrors the workflow/analysis model state from setup. Also anything shared across tasks just goes through the setup task (e.g. running a workflow). Some of this could be moved into a base class 
- A lot of tests have been moved around, depending on where the old functionality ended up
- I added a toolbar. Probably should have been in a different PR really..

I'd say this PR is working but (very!) incomplete, I just don't want to get into a situation where I try to add too much in one go.

I originally had most of the things in setup task at application level, but that would have stopped people from having multiple workflows open within the same application (I think!).

Also, is there a general idea that Tasks should be able to work independently? As in, the application should run in some form if we only had a results task and no setup (or the opposite). It would be no trouble to write it that way, but I don't know if there'd ever be a situation where you don't have both tasks.


force_wfmanager/tests/fixtures/evaluation-4.json
```diff
@@ -0,0 +1,259 @@
+{
```
jjenthought: A file to test if the workflow manager loads stuff properly

force_wfmanager/wfmanager.py
```diff
@@ -47,8 +61,15 @@ def close(self):
         fails, the application is not closed so he has a chance to try to
         save again. Overrides close from pyface.tasks.task_window """
 
-        # The attached wfmanager_task for saving methods
-        wfmanager_task = self.tasks[0]
+        # The attached wfmanager_setup_task for saving methods
```
jjenthought: This will not be needed when we have a BaseTask class for the common actions of the Setup & Result tasks. But that's for another PR maybe!

force_wfmanager/central_pane/result_table.py
```diff
@@ -37,6 +37,10 @@ class ResultTable(HasStrictTraits):
         ))
     )
 
+    def __init__(self, analysis_model):
```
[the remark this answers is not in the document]
dpinte: One issue with this is that you're breaking the standard traits initialization. I would strongly suggest using keyword arguments for your __init__ if you want to keep it. Doing it will allow you to remove the __init__ method without breaking things.

Make sense? 

Then you could even remove line 42 and just pass the kwargs to the `__init__` call.

force_wfmanager/central_pane/setup_pane.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(SetupPane, self).__init__(*args, **kwargs)
+
+    def _console_ns_default(self):
+        namespace = {
+            "task": self.task
+        }
+        try:
```
dpinte: this smells a bit. 

First, why do you need the `app` in the console? Is it for debugging? Second, you should probably rethink this and just contribute items to the console. There is a plugin supporting that in envisage IIRC.

This should not hold up this PR but ...

force_wfmanager/left_side_pane/results_pane.py
```diff
@@ -0,0 +1,50 @@
...
+from traitsui.api import View, UItem, VGroup
+
+from force_bdss.api import IFactoryRegistryPlugin
+from force_wfmanager.central_pane.result_table import ResultTable
+from force_wfmanager.central_pane.analysis_model import AnalysisModel
+
+
+class ResultsPane(TraitsDockPane):
+    """ Side pane which contains the WorkflowSettings (Tree editor for the
+    Workflow) and the Run button """
```
dpinte: Missing empty line after the docstring?

force_wfmanager/left_side_pane/results_pane.py
```diff
@@ -0,0 +1,50 @@
...
+
+    #: The analysis model containing the results
+    analysis_model = Instance(AnalysisModel)
+
+    traits_view = View(VGroup(
+        UItem('result_table', style='custom'),
+    ))
+
+    def __init__(self, analysis_model, *args, **kwargs):
+        self.analysis_model = analysis_model
```
dpinte: by initializing `analysis_model` before calling the Traits init machinerie, you might be breaking validation, init, etc.

I would strongly suggest going ahead with kwargs only if there is no strong reason to do it another way. Just rely on the traits constructor:

dpinte:
```
def __init__(self, analysis_model, *args, **kwargs):
￼         self.analysis_model = analysis_model
         super(ResultsPane, self).__init__(*args, **kwargs)
```

should at least be:

```
def __init__(self, analysis_model, *args, **kwargs):
         super(ResultsPane, self).__init__(*args, **kwargs)
￼         self.analysis_model = analysis_model
```

force_wfmanager/left_side_pane/results_pane.py
```diff
@@ -0,0 +1,50 @@
...
+    movable = False
+
+    #: Make the pane visible by default
+    visible = True
+
+    #: The factory registry containing all the factories
+    factory_registry = Instance(IFactoryRegistryPlugin)
+
+    #: The table displaying the results
+    result_table = Instance(ResultTable)
```
dpinte: any reason why this is not a Property that depends on `analysis_model` ?

force_wfmanager/left_side_pane/results_pane.py
```diff
@@ -0,0 +1,50 @@
...
+    floatable = False
+
+    #: Remove the possibility to move the pane in the GUI
+    movable = False
+
+    #: Make the pane visible by default
+    visible = True
+
+    #: The factory registry containing all the factories
+    factory_registry = Instance(IFactoryRegistryPlugin)
```
dpinte: what is this used for? No code depends on it. Why do we need it here?

force_wfmanager/wfmanager.py
```diff
@@ -1,16 +1,27 @@
...
+from pyface.tasks.api import TaskWindowLayout
+
+log = logging.getLogger(__name__)
 
 
 class WfManager(TasksApplication):
     id = 'force_wfmanager.wfmanager'
     name = 'Workflow Manager'
 
+    def __init__(self, plugins):
```
dpinte: If you want to keep an __init__ method here, please use keyword arguments. Something like 

```
    def __init__(self, plugins=None):
```

If you remove your __init__, then the traits init will still work!!! Right now, it won't.

```
plugins = [MyPlugin]
manager = WfManager(plugins) # works with your init but not traits init
manager = WfManager(plugins=plugins) # with with Traits but not yours 
```

force_wfmanager/wfmanager_plugin.py
```diff
@@ -1,35 +1,77 @@
...
+    TASK_EXTENSIONS = 'envisage.ui.tasks.task_extensions'
+
     id = 'force_wfmanager.wfmanager_plugin'
     name = 'Workflow Manager'
-    workflow_file = None
 
     tasks = List(contributes_to=TASKS)
+    task_extensions = List(contributes_to=TASK_EXTENSIONS)
+
+    def __init__(self, workflow_file):
```
dpinte: Same comment as before.

force_wfmanager/wfmanager_plugin.py
```diff
@@ -1,35 +1,77 @@
...
+            factory_registry=factory_registry,
+        )
         if self.workflow_file is not None:
-            wf_manager_task.open_workflow_file(self.workflow_file)
+            wf_manager_setup_task.open_workflow_file(self.workflow_file)
+
+        return wf_manager_setup_task
+
+    def _create_results_task(self):
+        factory_registry = self.application.get_plugin(
```
dpinte: You should decide which convention you use for ending parenthesis. Line 64 adds no line break while line 66 has a line break. I typically add a line break. I suggest you check what the standard is in the force repositories.

force_wfmanager/wfmanager_results_task.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class WfManagerResultsTask(Task):
+    """Task responsible for running the Workflow and displaying the results."""
+
+    id = 'force_wfmanager.wfmanager_results_task'
+    name = 'Results'
+
+    #: Workflow model.
+    workflow_m = Instance(Workflow, allow_none=False)
```
dpinte: `worfklow_m` is not very good as a name. Why not using `workflow_model`?

force_wfmanager/wfmanager_results_task.py
```diff
@@ -0,0 +1,215 @@
...
+
+    #: Analysis model. Contains the results that are displayed in the plot
+    #: and table
+    analysis_m = Instance(AnalysisModel, allow_none=False)
```
dpinte: Same comment.

force_wfmanager/wfmanager_results_task.py
```diff
@@ -0,0 +1,215 @@
...
+
+    #: Registry of the available factories
+    factory_registry = Instance(IFactoryRegistryPlugin)
```
dpinte: This trait does not seem to be used in this class

force_wfmanager/wfmanager_results_task.py
```diff
@@ -0,0 +1,215 @@
...
+
+    #: Side Pane containing the tree editor for the Workflow and the Run button
+    side_pane = Instance(ResultsPane)
+
+    #: The menu bar for this task.
+    menu_bar = Instance(SMenuBar)
+
+    #: The tool bar for this task.
```
dpinte: tool bar or tool bars?

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -40,95 +40,255 @@ class WfManagerTask(Task):
...
+    ui_hooks_managers = List(BaseUIHooksManager)
 
-    #: Flag which says if the menus should be disabled or not
-    _menu_enabled = Bool(True)
+    task_group = Instance(TaskToggleGroupAccelerator)
+
+    #: Results Task
+    results_task = Instance(Task)
+
+    def __init__(self, factory_registry, *args, **kwargs):
```
dpinte: usual comment about traits initializer

force_wfmanager/central_pane/setup_pane.py
jjenthought: I think it was already passed as a keyword argument anyway, so the __init__ was definitely unneccesary! It would be good to know how the traits initialisation works though. I was working on the assumption that traits magic would happen during `super(TraitSubclass, self).__init__()` but maybe this isn't completely true? Can you think of anything to watch out for if I'm doing this sort of thing?
For the console, it's just there for debugging and will eventually only be displayed in the applications 'debug' mode. I'm not too familiar with the details of how the console is set up though.

force_wfmanager/left_side_pane/results_pane.py
jjenthought: Well it looks like me and Stefano did this for a lot of the docstrings! I'll fix them as I see them

force_wfmanager/left_side_pane/results_pane.py
jjenthought: Yep, and as with the first comment we don't need init anyway!

force_wfmanager/wfmanager_results_task.py
jjenthought: Yeah good point! I think we'd just got too used to calling them that!

PR #241, opened 2018-09-17
Modal Dialogue Moved To Setup Pane

jjenthought:
Follow-on from https://github.com/force-h2020/force-wfmanager/pull/240. This moves the NewModalDialog functionality to the SetupPane, instead of opening a new window.


force_wfmanager/central_pane/setup_pane.py
```diff
@@ -1,25 +1,118 @@
...
+                    ),
+                    label="New Model Details",
+                    visible_when="selected_factory_name != 'None'",
+                    show_border=True,
+
+                ),
 
-    def __init__(self, *args, **kwargs):
-        super(SetupPane, self).__init__(*args, **kwargs)
+                # TODO:
```
jjenthought: I sometime uncomment this for debug things still, so it would be helpful to keep it around for now. As it says, it will be replaced in the course of the UI redesign

force_wfmanager/central_pane/setup_pane.py
```diff
@@ -31,3 +124,71 @@ def _console_ns_default(self):
             namespace["app"] = None
 
         return namespace
+
+    # Properties
+
```
jjenthought: If and object does not have a 'custom' view, we don't want to show the default one as there are a lot of traits which are not intended to be user-facing. I imagine there may well be a cleaner way to check for this though

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -32,40 +35,22 @@
...
 new_kpi_action = Action(name='New KPI...', action='new_kpi')
-delete_kpi_action = Action(name="Delete", action='delete_kpi')
 new_layer_action = Action(name="New Layer...", action='new_layer')
-delete_layer_action = Action(name='Delete', action='delete_layer')
-new_data_source_action = Action(name='New DataSource...',
-                                action='new_data_source')
-delete_data_source_action = Action(name='Delete', action='delete_data_source')
-edit_data_source_action = Action(name='Edit...', action='edit_data_source',
-                                 enabled_when=call_modelview_editable)
+
```
jjenthought: The delete context menu items are retained, along with addition actions with only add generic items (KPIs, Execution Layers)

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -235,131 +111,342 @@ class WorkflowTree(ModelView):
...
+                # Root node "Workflow"
+                TreeNodeWithStatus(
+                    node_for=[WorkflowModelView],
+                    auto_open=True,
+                    children='',
+                    name='Workflow',
+                    label='=Workflow',
+                    view=no_view,
+                    menu=no_menu,
+                    on_select=self.show_workflow_status
```
jjenthought: show_workflow_status is not implemented in this pr

force_wfmanager/left_side_pane/workflow_tree.py
```diff
@@ -235,131 +111,342 @@ class WorkflowTree(ModelView):
...
+        """Assigns the correct add/remove entity functions based on which
+        object is selected in the workflow tree"""
+        instance_select_function = {
+            'Notification Listeners': self.delete_notification_listener,
+            'MCO': self.delete_mco, 'KPIs': self.delete_kpi,
+            'Execution Layers': self.delete_layer,
+            'DataSources': self.delete_data_source,
+            'Parameters': self.delete_parameter
+        }
+        factory_select_function = {
```
jjenthought: It's kind of annoying having to have this, but the requirements on selecting different factories are just about different enough that having a generic function would be harder to read.

PR #244, opened 2018-10-02
Scrollable Error Message

jjenthought:
When there are multiple error messages, only display one at a time but with the ability to navigate through the different messages.


jjenthought:
Don't think this is really needed

PR #252, opened 2019-03-18
Bumped Python version to 3.6

ml-evs:
To follow corresponding changes to force-bdss and the plugins.

Removed the explicit requirement for traits, as the relevant version will be pulled in by force-bdss anyway.

### DIFF
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -1,8 +1,8 @@
 import click
 from subprocess import check_call
 
-DEFAULT_PYTHON_VERSION = "3.5"
-PYTHON_VERSIONS = ["3.5"]
+DEFAULT_PYTHON_VERSION = "3.6"
+PYTHON_VERSIONS = ["3.6"]
 
 ADDITIONAL_CORE_DEPS = [
     "pyface==6.0.0-1",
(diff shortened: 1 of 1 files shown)


ci/__main__.py
```diff
@@ -1,17 +1,16 @@
...
+PYTHON_VERSIONS = ["3.6"]
 
 ADDITIONAL_CORE_DEPS = [
-    "pyface==6.0.0-1",
+    "pyface==6.0.0-2",
     "pygments==2.2.0-1",
     "pyqt==4.11.4-7",
     "qt==4.8.7-7",
     "sip==4.17-4",
-    "traits==4.6.0-1",
```
jjenthought: I'm not actually sure what the standard way to do this is, but could we ask for traits 5.0.0-1 in this list? It gets pulled in as a dependency right now, but I think that may mean it gets the latest one by default. Which I'd imagine is fine almost all the time, but could jump in as a 'silent' error if a newer version of traits breaks something in future

ml-evs: I guess my thinking was that force_bdss is an unlisted dependency for this, which will have bundled the version of Traits that it needs. Perhaps we should include force_bdss here instead?

jjenthought: That would be nice! But I think 'cross-linking' the build scripts might be harder to maintain at this stage since they don't have proper edm/pip egg distributions. I really don't know much about this stuff though.
I think traits 5 actually gets pulled in as a dependency of envisage in force_bdss in any case!
So I'd be happy to keep the deletion, I guess also the less requirements you put on edm the easier it is for it to resolve the dependencies!

PR #253, opened 2019-03-18
Set window resolution at command line

jjenthought:
Adds a command line flag `--window size WIDTH HEIGHT` allowing the window size to be set. This should be helpful for simphony-remote stuff where the resolution of the browser window could be fairly small!


ml-evs:
This PR Works For Me™ on macOS Mojave, though I have to delete the saved GUI state under `$HOME/.enthought/force_wfmanager.wfmanager/tasks/application_memento` for it to accept the `--window-size` arg. Happy to accept if this is useful for simphony-remote as is.

PR #255, opened 2019-03-20
Improved ResultsPane performance with TabularEditor and enabled multi-point selections

### LINKED ISSUE #254: Slowdown when plotting any moderately large number of points
If we run a workflow which outputs ~100+ result points we get a lot of slowdown which renders the UI unusable. This seems to get worse as more points are added, suggesting there is some O(n^2) operation being repeatedly performed.
But this needs diagnosing properly!

ml-evs:
This PR fixes #254 by switching the results view from TableEditor to a TabularEditor. This had a few side effects, as listed below:

- A fix to TraitsUI was required enthought/traitsui#521 which has since been merged to master, and thus this PR pins the TraitsUI version required by this package to that particular commit, to be switched back in the future (see #258)
- By switching to TabularEditor, you can now select multiples rows or multiple points on the plot to be highlighted.
- The plotting style was modified slightly to accommodate multiple selections.
- One other minor change is that the Force logo now rescales depending on the window size, rather than enforcing a minimum width ~500 pixels.
- Added --profile flag to force_wfmanager entrypoint that provides .prof/.pstats files.

### DIFF
--- a/force_wfmanager/central_pane/analysis_model.py
+++ b/force_wfmanager/central_pane/analysis_model.py
@@ -72,11 +72,12 @@ class AnalysisModel(HasStrictTraits):
         return self._selected_step_indices
 
     def _set_selected_step_indices(self, values):
-        if values:
-            self._selected_step_indices = values
+        self._selected_step_indices = values
+        if values is not None:
             for value in values:
                 if (isinstance(value, int) and not
                         (0 <= value < len(self._evaluation_steps))):
+                    self._selected_step_indices = None
                     raise ValueError(
                         "Invalid value for selection index {}. Current "
                         "number of steps = {}".format(
--- a/force_wfmanager/central_pane/analysis_model.py
+++ b/force_wfmanager/central_pane/analysis_model.py
@@ -72,12 +72,14 @@ class AnalysisModel(HasStrictTraits):
         return self._selected_step_indices
 
     def _set_selected_step_indices(self, values):
-        self._selected_step_indices = values
+        """ Check the requested indices of selected rows, and use the
+        requested values if they exist in the table, or are None.
+
+        """
         if values is not None:
             for value in values:
                 if (isinstance(value, int) and not
                         (0 <= value < len(self._evaluation_steps))):
-                    self._selected_step_indices = None
                     raise ValueError(
                         "Invalid value for selection index {}. Current "
                         "number of steps = {}".format(
@@ -85,6 +87,8 @@ class AnalysisModel(HasStrictTraits):
                             len(self._evaluation_steps)
                         )
                     )
+
+        self._selected_step_indices = values
 
     def clear(self):
         """ Sets :attr:`value_names` to be empty, removes all entries in the
--- a/force_wfmanager/central_pane/analysis_model.py
+++ b/force_wfmanager/central_pane/analysis_model.py
@@ -20,7 +20,7 @@ class AnalysisModel(HasStrictTraits):
 
     #: Selected step, used for highlighting in the table/plot.
     #: Listens to: :attr:`value_names`
-    _selected_step_index = Either(None, Int())
+    _selected_step_indices = Either(None, List(Int()))
 
     # ----------
     # Properties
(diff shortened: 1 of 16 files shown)


force_wfmanager/central_pane/result_table.py
```diff
@@ -62,8 +49,25 @@ def _get_columns(self):
                 for index, name
                 in enumerate(self.analysis_model.value_names)]
 
-    # Response to model change
+    # ----
+    # View
+    # ----
+    def traits_view(self):
```
rkern:
```suggestion
    def default_traits_view(self):
```

`traits_view()` has a number of functions. The method to implement to provide a default `View` programmatically is `default_traits_view()`.

ml-evs: Fixed, thanks.

force_wfmanager/central_pane/analysis_model.py
```diff
@@ -68,31 +68,34 @@ def add_evaluation_step(self, evaluation_step):
...
-                    value,
-                    len(self._evaluation_steps)
-                )
-            )
-
-        self._selected_step_index = value
+    def _get_selected_step_indices(self):
+        return self._selected_step_indices
+
+    def _set_selected_step_indices(self, values):
```
jjenthought: Is there any danger in making the assignment before the checks? I spotted a attr (`_plot_index_datasource`) in plot.py which listens to `selected_step_indices`, not sure if that would get updated immediately though.
I guess my thought is that if values contains a 'missing' index it might fail on the chaco side of things?

force_wfmanager/central_pane/plot.py
```diff
@@ -60,9 +61,9 @@ class Plot(HasStrictTraits):
     _plot_data = Instance(ArrayPlotData)
 
     #: Datasource of the plot (used for selection handling)
-    #: Listens to: :attr:`analysis_model.selected_step_index
+    #: Listens to: :attr:`analysis_model.selected_step_indices
     #: <force_wfmanager.central_pane.analysis_model.AnalysisModel.\
```
jjenthought: If you're familiar with sphinx/rst and know of any way to shorten these long function references but still have the docs link to it that would be really good! As you can see I couldn't figure out a way to do it haha

force_wfmanager/central_pane/plot.py
```diff
@@ -111,22 +112,28 @@ def __plot_default(self):
...
             scatter_plot,
             threshold=10,
-            selection_mode="single",
+            multiselect_modifier=KeySpec(None, "shift"),
+            selection_mode="multi",
         ))
         overlay = ScatterInspectorOverlay(
             scatter_plot,
-            hover_color="blue",
+            hover_color=(0, 0, 1, 1),
```
jjenthought: Is this tuple R,G,B,alpha? If it is would it make sense to have the hover selection be transparent and the actual selection be solid? Not exactly a big deal either way..

force_wfmanager/central_pane/tests/test_analysis_model.py
```diff
@@ -34,35 +34,41 @@ def test_change_selection(self):
         self.analysis.add_evaluation_step((3, 4))
         self.analysis.add_evaluation_step((5, 6))
 
-        self.analysis.selected_step_index = 0
-        self.analysis.selected_step_index = 2
-        self.analysis.selected_step_index = None
+        self.analysis.selected_step_indices = [0]
```
jjenthought: Maybe we could throw a `self.analysis.selected_step_indices = [0, 1]` in here? I mean there's no specific result to test against, but just to check there's no knock-on effects when selecting multiple points.

jjenthought: Could also try `.._indices = [0,3]` and check it raises ValueError

force_wfmanager/central_pane/analysis_model.py
ml-evs: I think you're right, though it's hard to test. I've refactored it so that `selected_step_indices` only gets modified if the `values` passed to `_set_selected_step_indices` are valid.

force_wfmanager/central_pane/plot.py
ml-evs: No idea I'm afraid! Google wasn't too helpful either.

force_wfmanager/central_pane/plot.py
ml-evs: I think this way round works, you can see explicitly which point you're about to select when you're hovering (if there are lots of points in close proximity), then then once selected, you get a larger marker that still lets you see nearby points, and their colour (for e.g. the next PR).

force_wfmanager/central_pane/plot.py
corranwebster: You can put as much of the tail as you need to be unambiguous, starting with a `.`, (and using `~` in front makes names prettier in actual docs, so
```
:attr:`~.selected_step_indices`
```
will probably work.  If there is ambiguity then Sphinx will raise a warning.

PR #257, opened 2019-03-26
Added colour mapped scatter plots

ml-evs:
Adds option to colour points by a third column in results pane, with choice of continuous/discrete colourmaps.

Current issues:
- ~GUI displays cmaps selection as `<function $name #memaddr>`, rather than just the name.~
  - solved
- ~When switching colourmaps, the plot axes reset (losing e.g. any existing zoom/pan).~
  - solved
- ~Discrete colourmaps cause a crash when used with e.g. floating point data.~
  - disabled discrete colormaps for now

### DIFF
--- a/force_wfmanager/central_pane/plot.py
+++ b/force_wfmanager/central_pane/plot.py
@@ -1,7 +1,8 @@
 from chaco.api import ArrayPlotData, ArrayDataSource, ScatterInspectorOverlay
-from chaco.api import Plot as ChacoPlot, ScatterPlot
+from chaco.api import Plot as ChacoPlot
+from chaco.api import BaseXYPlot, ColormappedScatterPlot
 from chaco.default_colormaps import (
-    color_map_functions, discrete_color_map_functions, viridis,
+    color_map_functions, discrete_color_map_functions, viridis, Dark2
 )
 from chaco.tools.api import PanTool, ScatterInspector, ZoomTool
 from enable.api import Component, ComponentEditor
@@ -40,13 +41,12 @@ class Plot(HasStrictTraits):
     #: Optional choice of colormap
     colormap_type = Enum('Continuous', 'Discrete')
     colormap = Enum(values='_available_colormaps',
-                    depends_on='_available_colormaps[]')
+                    depends_on='_available_colormaps')
     color_plot = Bool(False)
 
     #: Button to reset plot view. The button is active if :attr:`reset_enabled`
     #: is *True* and inactive if it is *False*.
     reset_plot = Button('Reset View')
-    _scatter = Instance(ScatterPlot)
 
     # --------------------
     # Dependent Attributes
@@ -65,6 +65,9 @@ class Plot(HasStrictTraits):
     )
     __discrete_colormaps = List(discrete_color_map_functions)
     _available_colormaps = __continuous_colormaps
+
+    #: Exposes particular attributes of the plot for tweaking
+    _axis = Instance(BaseXYPlot)
 
     #: A local copy of the analysis model's value names
     #: Listens to: :attr:`analysis_model.value_names
@@ -131,7 +134,7 @@ class Plot(HasStrictTraits):
     def __plot_default(self):
         return self.plot_scatter()
 
-    @on_trait_change('color_plot')
+    @on_trait_change('color_plot,colormap')
     def change_plot_style(self):
         if self.color_plot:
             self._plot = self.plot_cmap_scatter()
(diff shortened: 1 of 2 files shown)


ml-evs:
This PR also contains the changes of #255 , so may need to do some rebasing to untangle them before merging...

force_wfmanager/central_pane/plot.py
```diff
@@ -77,28 +125,50 @@ class Plot(HasStrictTraits):
...
     def __init__(self, analysis_model, *args, **kwargs):
         self.analysis_model = analysis_model
         super(Plot, self).__init__(*args, **kwargs)
 
         self.update_data_arrays()
 
-    # Defaults
+    # Disable discrete color maps for now
+    @on_trait_change('colormap_type')
+    def update_colormap_list(self):
```
jjenthought: We could remove the commented code and make an issue for this with the commented stuff there. There's also the thing of currently having a one item dropdown menu now we took out the discrete maps. 
I'm never sure whether it's better to leave in things that are going to be added to in future or take them out and have each PR be self-contained. I'd be happy to leave the dropdown as it is, maybe it'll be good as a reminder to add the discrete maps at some point?

force_wfmanager/central_pane/plot.py
```diff
@@ -43,6 +63,34 @@ class Plot(HasStrictTraits):
...
     _plot = Instance(Component)
 
+    #: Possible colormaps given color_map type
+    #: Listens to :attr:`colormap_type` and relies on default being the
+    #: The default is set by the first entry of this list.
+    __continuous_colormaps = Dict(
+        {cmap.__str__().split()[1]: cmap
+         for cmap in color_map_functions}
+    )
+    __discrete_colormaps = Dict(
```
jjenthought: I think you did mention it already, but what was the issue with the discrete maps? Was it that they crashed for non int values?
If it seemed like it would be quite a difficult thing to fix, I'd remove the `__discrete_colormaps` and associated attrs for now. But if not we can leave it in and wait for the fix before merging this

force_wfmanager/central_pane/plot.py
```diff
@@ -1,11 +1,15 @@
...
+from chaco.default_colormaps import (
+    color_map_functions, discrete_color_map_functions
+)
 from chaco.tools.api import PanTool, ScatterInspector, ZoomTool
 from enable.api import Component, ComponentEditor
 from enable.api import KeySpec
 from traits.api import (
     Button, Bool, Enum, HasStrictTraits, Instance, List, Property, Tuple,
-    on_trait_change
+    on_trait_change, Dict,
```
jjenthought: I think alphabetising (sp?) imports are either an enthought style thing, or possibly just something Mark told me to do once! It does help prevent git making merge conficts so it's not just pedantry haha

force_wfmanager/central_pane/plot.py
ml-evs: Basically, yeah. If you try to use a discrete colourmap for floats, chaco trashes the plot. One possible way around this would be to mark certain columns in the analysis model as "categorical", but it might be more effort than its worth.

force_wfmanager/central_pane/plot.py
ml-evs: I agree (cf. conversation below), it might be too much of a pain to support this and would certainly complicate the analysis model.

ml-evs:
Regarding the colourbar, I've got the code for it, but as chaco treats colourbars as separate plots, `self._plot` has to become an `HPlotContainer` so all the of the resize_plot/overlay methods would qneed to be rewritten. There's also some complication with updating the colourbar dynamically as the colourmap changes. I'm happy to delay this PR until I've had a proper go at doing this.

force_wfmanager/central_pane/plot.py
ml-evs: That's a good tip, thanks!

jjenthought:
> Regarding the colourbar, I've got the code for it, but as chaco treats colourbars as separate plots, `self._plot` has to become an `HPlotContainer` so all the of the resize_plot/overlay methods would qneed to be rewritten. There's also some complication with updating the colourbar dynamically as the colourmap changes. I'm happy to delay this PR until I've had a proper go at doing this.
> 
> (See e.g. https://github.com/enthought/chaco/blob/master/examples/demo/basic/zoomable_colorbar.py)

I think that's a significant enough change to get it's own issue/PR! If it needs a few things to be re-written it's nice to have that all tied to a single point when looking back through the commit history. If you're still ok to merge this I'd definitely go ahead with it

PR #264, opened 2019-04-10
Minor interface fixes

### LINKED ISSUE #261: Text in "Available Plugins" Listbox is Editable
They clearly should be read-only.

Screen Shot 2019-04-01 at 10 32 57 am

ml-evs:
Some minor fixes to the wfmanager interface.

- Changes HTML font to sans-serif to match the rest of the GUI.
- Use trait `label` rather than `name` in model descriptions if defined.
- Fixes #261 by making plugin list uneditable.


jjenthought:
Or even right now! I thought it was going to restart the ci but apparently not!

PR #269, opened 2019-04-24
Save and load results alongside the workflow that created them

### LINKED ISSUE #266: Ability to save results table to file
Users should be able to save the results of their simulations to a file after the workflow is complete, so they can be analysed elsewhere if required.

There should probably also be a way to save the results alongside the workflow that created them, and load them back in when requested.

ml-evs:
Closes #266 .

This PR adds the following functionality to the workflow manager and its GUI:

- Export analysis model (i.e. table of results) to JSON or CSV, for use elsewhere.
- Save "Project" files containing the analysis model *and* the workflow used to generate it.
    - This required a change to the results task such that it takes a copy of the workflow when it has been run, and does not update with the setup task's copy.
- Open project files from the results view, loading the correct state into both results and setup task, displaying the results in the default view.

### DIFF
--- a/force_wfmanager/central_pane/analysis_model.py
+++ b/force_wfmanager/central_pane/analysis_model.py
@@ -139,7 +139,7 @@ class AnalysisModel(HasStrictTraits):
 
         """
         json_representation = {}
-        for ind, name in enumerate(self.value_names):
+        for name in self.value_names:
             json_representation[name] = []
         for step in self.evaluation_steps:
             for ind, name in enumerate(self.value_names):
--- a/force_wfmanager/central_pane/analysis_model.py
+++ b/force_wfmanager/central_pane/analysis_model.py
@@ -117,19 +117,17 @@ class AnalysisModel(HasStrictTraits):
         self._clear_evaluation_steps()
 
     def from_dict(self, data):
-        """ Load the :attr:`value_names` and :attr:`evaluation_steps`
-        from a dictionary.
+        """ Delete all current data and load :attr:`value_names` and
+        :attr:`evaluation_steps` from a dictionary.
 
         """
+        self._clear_evaluation_steps()
         self.value_names = tuple(key for key in data.keys())
-        steps = []
-        for ind, values in enumerate(data[self.value_names[0]]):
+        for ind, _ in enumerate(data[self.value_names[0]]):
             step = []
             for column in self.value_names:
                 step.append(data[column][ind])
-            steps.append(step)
-
-        self._evaluation_steps = steps
+            self.add_evaluation_step(step)
 
     def as_json(self):
         """ Returns a JSON representation with column names as keys and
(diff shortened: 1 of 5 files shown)


ml-evs:
I don't think I can improve the coverage without going overboard, so this is probably ready to review.

force_wfmanager/wfmanager_results_task.py
```diff
@@ -178,11 +190,224 @@ def _default_layout_default(self):
...
+                project_json = json.load(fp)
+
+                # share the analysis model with the setup_task
+                self.analysis_model.from_dict(project_json['analysis_model'])
+                self.setup_task.analysis_model = self.analysis_model
+
+                # create two separate workflows, so that setup task can be
+                # edited without changing the results task copy
+                reader = WorkflowReader(self.setup_task.factory_registry)
+                fp.seek(0)
```
jjenthought: Is this the part where it wouldn't copy things in a straightforward way?

force_wfmanager/central_pane/analysis_model.py
```diff
@@ -103,3 +115,80 @@ def clear_steps(self):
...
+        """ Returns a JSON representation with column names as keys and
+        values stored as lists under those keys.
+
+        Returns
+        -------
+        dict: a dictionary containing the JSON representation.
+
+        """
+        json_representation = {}
+        for ind, name in enumerate(self.value_names):
```
jjenthought: This one doesn't have to be an enumerate from the looks of things

force_wfmanager/central_pane/tests/test_analysis_model.py
```diff
@@ -82,3 +83,67 @@ def test_clear(self):
...
+
+        self.assertTrue(success)
+
+    def test_write_to_csv(self):
+        json_rep = {'foo': [1, 3, 5],
+                    'bar': [2, 4, 6],
+                    'label': ['x', 'y', 'z']}
+        self.analysis.from_dict(json_rep)
+        self.assertEqual(self.analysis.as_json(), json_rep)
+        mock_open = mock.mock_open()
```
jjenthought: neat I didn't know that mock had this! It also made me very confused since I looked at L139 onwards first

force_wfmanager/wfmanager_results_task.py
ml-evs: Yep, think this is the best way of doing it, rather than hacking __deepcopy__!

force_wfmanager/central_pane/tests/test_analysis_model.py
ml-evs: It's quite neat! Unfortunately `json.dump` doesn't write line-by-line so I can't do the write call comparison as in the csv example. Probably overkill anyway, and `assert_any_call` doesn't let you enforce an order!

PR #277, opened 2019-05-30
Fixes for windows build

jjenthought:
There was an error in `setup.py` where `package_data` was given as a string rather than a list of strings. This was fine on unix but I guess on windows it tries to do some path conversion and fails there.
Also upgraded numpy & chaco since we need them for plugins using scipy (which is most of them!)


flongford:
The build failing looks due to the absence of `mock` - https://github.com/force-h2020/force-wfmanager/pull/278 will fix this

PR #286, opened 2019-06-13
Fixes to comply to Flake8 3.7.7

nicolasap-dm:
I'm upgrading Flake8 in `force-bdss`, and that will require these updates here.

### DIFF
--- a/force_wfmanager/server/zmq_server.py
+++ b/force_wfmanager/server/zmq_server.py
@@ -229,13 +229,13 @@ class ZMQServer(threading.Thread):
         self.ports = None
         try:
             self._pub_socket.close()
-        except:
+        except Exception:
             pass
         self._pub_socket = None
 
         try:
             self._sync_socket.close()
-        except:
+        except Exception:
             pass
         self._sync_socket = None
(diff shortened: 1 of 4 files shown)


force_wfmanager/server/zmq_server.py
```diff
@@ -229,13 +229,13 @@ def _close_network_sockets_noexc(self):
         self.ports = None
         try:
             self._pub_socket.close()
-        except:
+        except Exception:
```
nicolasap-dm: Some would argue this is ugly. But I'm not changing behavior here (and this fits the docstring "_This method throws away all exceptions that the operation might encounter_")

flongford: Agreed: _Explicit is better than implicit._

nicolasap-dm: (to be clear: the change was dictated by Flake8, I just felt the need to justify myself for having used an _equally ugly_ `except`, where every exception -- including, I believe, deliberate keyboard interrupts -- is caught without scrutiny)

PR #288, opened 2019-06-14
Upgrade dependencies with adaptations for traitsui and pyface 6.1.x

nicolasap-dm:
This PR upgrades all the explicit dependencies, including traitsui and pyface 6.1.x.

Interaction with the new libraries caused test errors, especially in `pyface.ui.qt4.util.gui_test_assistant.GuiTestAssistant.tearDown`. Now all the relevant applications are closed before reaching `tearDown`, or are just closed in other test classes which had problems cleaning up the UIs.


ci/__main__.py
```diff
@@ -5,23 +5,18 @@
...
 
 ADDITIONAL_CORE_DEPS = [
-    "pyface==6.0.0-2",
+    "pyface==6.1.0-2",
     "pygments==2.2.0-1",
-    "pyqt==4.11.4-7",
-    "qt==4.8.7-7",
-    "sip==4.17-4",
-    "traitsui==6.0.0-2",
+    "pyqt==4.12.1-1",
```
corranwebster: Be careful of pyqt 4.12 - it has been buggy in the unittests for Pyface an TraitsUI where 4.11 has been comparatively stable.

nicolasap-dm: This suggestion might save my day. Checking!

nicolasap-dm: It doesn't seem to affect my case. Pushing anyway, because I have different set of test failures locally and in CI.

force_wfmanager/wfmanager.py
```diff
@@ -65,8 +65,10 @@ def _remove_tasks(self):
                 window.remove_task(task)
 
     # FIXME: If the underlying envisage TasksApplication function is fixed to
-    #        work correctly, this will not be needed
+    #        work correctly, this will not be needed.
     def create_window(self, layout, restore, **traits):
+        """ Creates a new TaskWindow.
+        """
```
nicolasap-dm: (This is not strictly related to this PR)

This was required by a warning in Sphinx, treated as fatal: https://travis-ci.com/force-h2020/force-wfmanager/builds/115790867#L772

and it was probably caused by the colon after `Parameters` here:

https://github.com/enthought/envisage/blob/01696d52807747c8917d486e3852acf166ae0076/envisage/ui/tasks/tasks_application.py#L171-L175

nicolasap-dm:
Yes, this can be reviewed, thanks

force_wfmanager/tests/test_wfmanager.py
```diff
@@ -23,47 +24,57 @@ def setUp(self):
         super(TestWfManager, self).setUp()
         self.wfmanager = ProbeWfManager()
 
+    @contextlib.contextmanager
```
flongford: Out of curiosity, is use of this library preferred when performing UI tests or just a work around for this particular case?

flongford:
Do you know whether the FIXME statements referring to `traitsui/qt4/ui_panel.py` in `wfmanager.py` are addressed in the library upgrades?

force_wfmanager/wfmanager.py
flongford: Do you know whether the FIXME statements referring to `traitsui/qt4/ui_panel.py` in `wfmanager.py` are addressed in the library upgrades?

force_wfmanager/tests/test_wfmanager.py
nicolasap-dm: I had originally created a "cleanup" method that would mirror `create_tasks`, in order to arrive at the `tearDown` stage "prepared". Since that cleanup had to be run for any test that uses `create_tasks`, and since a proper clean up was nice to have also if the test errored/failed, having it wrapped in a context manager helped.

force_wfmanager/wfmanager.py
nicolasap-dm: I haven't checked. Checking now.

force_wfmanager/tests/test_wfmanager.py
flongford: Cheers, makes sense - it looks neat, so I was wondering what other situations would it might be useful for.

nicolasap-dm: I'd say in every case where you have some cleanup tasks that are not completely general for the test class (which go in `tearDown`) and have to happen before `tearDown` (`addCleanup` happens after)

force_wfmanager/wfmanager.py
nicolasap-dm: I'm still not sure what was happening that required #228 (which introduced these FIXMEs). However, we haven't got any functional updates in https://github.com/enthought/traitsui/commits/master/traitsui/qt4/ui_panel.py, and removing the overrides fails this test:

https://github.com/force-h2020/force-wfmanager/blob/b7db3c397369bd2cc497eb06f5269e0f44b0c1ca/force_wfmanager/tests/test_wfmanager.py#L68-L72

flongford: Ok, cheers for checking. I can't seem to find an issue related to these FIXMEs - could you raise one referencing the library dependencies being updated so as to keep track of them in the future?

nicolasap-dm: Done: #300

nicolasap-dm:
(Working to make this PR work with the new `master`, as #293 adds a test that doesn't clean up things properly for this traitsui/pyface version)

PR #293, opened 2019-06-19
Discard corrupted state file

### LINKED ISSUE #290: A broken application_memento can lead to obscure test failures.
The `application_memento` file (that on Linux is at `~/.enthought/force_wfmanager.wfmanager/tasks/qt4`) can be found (probably as a result of faulty runs) in a state where subsequent launches of the application or of the test suite would fail.

The last time this has happened to me (#289), the errors I were facing in the test suite were similar to:

```
======================================================================
ERROR: test_switch_task (force_wfmanager.tests.test_wfmanager.TestWfManager)
----------------------------------------------------------------------
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/tests/test_wfmanager.py", line 126, in test_switch_task
    self.create_tasks()
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/tests/test_wfmanager.py", line 93, in create_tasks
    self.wfmanager.run()
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/envisage/ui/tasks/tasks_application.py", line 305, in _create_windows
    restore=self.always_use_default_layout)
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/wfmanager.py", line 72, in create_window
    layout, not restore, **traits
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/envisage/ui/tasks/tasks_application.py", line 223, in create_window
    layout = self._restore_layout_from_state(layout)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/envisage/ui/tasks/tasks_application.py", line 351, in _restore_layout_from_state
    match.active_task = layout.get_active_task()
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 236, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: The 'active_task' trait of a TaskWindowLayout instance must be a string, but a value of None <class 'NoneType'> was specified.
```

this test in particular has already proven to be sensible to the `memento` file (in a different way), since these line have been added at some point in the past:

https://github.com/force-h2020/force-wfmanager/blob/cc7c1c1c6e798caddd58045b4e68cd338566a8a7/force_wfmanager/tests/test_wfmanager.py#L125-L132

----

We should 

- decide whether this saved state is needed at all

and, if we decide to keep it, 

- make the application skip it if it's faulty (or provide an useful error message)

nicolasap-dm:
Closes #290 by removing a corrupted state file.

Alternative solutions in case of a corrupted file, for discussion:
- prevent launch and ask the user to delete the file and launch again
- do as this PR does but make a backup instead of removing the file

I don't think there's such a need to be that much cautious.


flongford:
Looks like a good work around.

The alternatives would seem to rely on the user knowing about the backend in order to be useful. Also, considering that there are only few ways for the user to customise the display (mainly window size) then the `application_momento` file seems a bit redundant.

PR #296, opened 2019-06-20
ENH: Verify Workflow on any labelled trait change in DataSourceModel

### LINKED ISSUE #280: VerifyWorkflowEvent only trigged under certain conditions in GUI
Currently a `VerifyWorkflowEvent` is only triggered when:

- A `DataSource` is added or removed from the workflow
- A `KPI` is added or removed from the workflow
- The `name` or `objective` of a `KPI` is changed
- An `ExecutionLayer` is added or removed from the workflow
- A `MCO` is added or removed from the workflow
- A `MCOParameter` is added or removed from the workflow
- The `name` or `type` of a `MCOParameter ` is changed
- A `NotificationListener` is added or removed from the workflow
- An `InputSlot` or `OutputSlot` is added or removed from the workflow
- The `name` of an `InputSlot` or `OutputSlot` is changed

Ideally we would like the workflow to be verified any time the user may make a change that could lead to the failure of the BDSS and therefore raise this before execution. For example, if any trait of a `DataSource` is entered incorrectly. Since these traits are variable between plugins, they are unknown to the WfManager before startup and therefore need to be listened to dynamically.

flongford:
This PR closes #280 

It also includes some small typo fixes in the `WorkflowInfo` class relating to the `selected_factory_name` attribute.

A `verify_workflow_event` will now occur when any trait of a `DataSourceModel` is changed in the wf_manager that has the metadata `verfiy=True`.

Previous work
--------------
Previously, additional complications occurred due to instantiation of the `TableRow` classes in `data_source_view.py`. This caused the default `traits_view` attribute on `DataSourceView` not to be assigned in the UI before it was called. Only if a `traits_view` attribute had been defined on the `DataSourceModel` prior to the relevant `DataSourceModelView` being initialised, would the error not arise.

This means either by explicitly defining a `traits_view` object in the `DataSourceModel` class constructor, or by calling the `DataSourceModel.trait_view()` method before binding to a `DataSourceModelView` (https://docs.enthought.com/traitsui/traitsui_user_manual/tips.html).
 
Originally, `model.trait_view()` was called in the `DataSourceModelView.__init__` method, before `self.model = model`. However, this was a fragile workaround and did not address the root cause of why a `traits_view` object needed to be called before binding, when there is already a default view (alphabetical order of trait names) associated to each `HasTraits` object?

Recent work
------------
Since #313, the `TableRow` class no longer instantiates a `model` attribute before calling `super().__init__()`, and so this error associated with `traits_view` no longer occurs. 

Consequently, we can add the change handler `model.+verify` to the `data_source_model_change` method of `DataSourceView` in order to also react whenever a trait with the metadata `verify` is changed.

This requires the developer to include this metadata in any additional plugin that is contributed to the WfManager, however it is a relatively simple addition and could be easily included in the `force-bdss` documentation.


force_wfmanager/ui/setup/setup_pane.py
```diff
@@ -299,6 +299,9 @@ def sync_selected_mv(self):
         self.selected_mv = self.task.side_pane.workflow_tree.selected_mv
         if self.selected_mv is not None:
             if isinstance(self.selected_mv.model, BaseModel):
+                #: FIXME - this will raise an AttributeError if
+                #: self.selected_mv.model has not initialised a traits_view
+                #: object
                 self.selected_model = self.selected_mv.model
```
flongford: If `DataSourceModelView` listens to all traits on a `DataSouceModel`, this returns:

`AttributeError: 'ProbeDataSourceModel' object has no attribute 'traits_view'`

For certain selected models.

flongford: This comment is now outdated

Issue #298, opened 2019-06-21
Refactoring of communication and dependence between traits needed


flongford:
Currently there are multiple ways of traits communicating with each other.

For example, differences in syntax using the `@triggers_verify` decorator in `workflow_tree.py` and the `@on_trait_change('object.verify_workflow_event')`  decorator in other modules.

This causes many redundant calls to getter and setter functions when objects are selected or changed, since it is easy to lose track of inter-trait dependencies.

flongford:
This issue is now outdated.

PR #299, opened 2019-06-24
Refactor of Property traits' getters to reduce redundant communication

flongford:
Addresses issue in #298

This PR removes the `Property` status from several traits that were being updated via `depends_on='object'` and instead uses the `on_trait_change('object')` on their getter method.

Refactoring communication in this way reduces the number of automated calls taking place, which had become large due to multiple intertwined dependencies. Some `Property` traits have become cached instead, though this has been used sparingly as it neglects to address the underlying issue of messy communication.

In doing so, the `selection` and `trigger_verify` decorators being using in `workflow_tree.py` have also been removed in order to make the code style more uniform. 

Also contains updates from #294 and #296 

## Further work:

Tidying up synchronisation of shared trait attributes between panes, trees and model views.

### DIFF
--- a/force_wfmanager/ui/review/plot.py
+++ b/force_wfmanager/ui/review/plot.py
@@ -171,7 +171,7 @@ class Plot(HasStrictTraits):
             marker_size=4,
             bgcolor="white")[0]
 
-        plot.set(title="Plot", padding=75, line_width=1)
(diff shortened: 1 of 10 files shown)


flongford:
This PR has now been incorporated into #299, and so is redundant

Issue #285, opened 2019-06-13
MCO Parameters and KPIs to share the same model view


flongford:
At the moment each MCO parameter and KPI has its own model view, which takes up hardly any space on the setup pane.

It would look better if they were all included in the same model view, as a group.

flongford:
This was implemented in #304 - ongoing issues relating to MCO parameters and kpis in the same view should reference this PR

Issue #282, opened 2019-06-12
ENH: Graphical Representation of Execution Layer Structure 


flongford:
At the moment the 'Execution Layer' model view displays redundant information. It would also be useful for the user to have a clearer understanding of how each layer and base data source fits together in a visual form.

This could be achieved by embedding an enable/kiva panel displaying each data source as a flow diagram-like object with arrows pointing to input/output variables being passed between them.

To begin with the representation could be static, though future adaptions could make this panel dynamic and therefore augment the current text-based UI structure.

flongford:
Creating a directed graph object of the workflow using `networkx` would make this representation straight forward to display.

flongford:
If we look to redesign the current prototype, we should probably look to use the ETS package `graphcanvas` https://github.com/enthought/graphcanvas

PR #304, opened 2019-07-11
Refactoring Tree Editor / MCO Parameter & KPI views

flongford:
This PR contains a mixture of elements from #285, #298 and #299. It also builds on previous refactoring in #287 to separate the GUI workflow into 3 distinct phases:

Build Hierarchical Process ---> Optimize Process ---> View Results

In doing so, it should become easier for a user to construct workflows with the UI, rather than inputting a JSON script.

Work Done
-----------
Extensive refactoring to the setup task has taken place, with the following results:

1. `TreePane` object is now called `SidePane` (with `tree_pane.py` renamed to `side_pane.py`)
2. A new object `SystemState` holds temporary information regarding the current user selection in order to act as a conduit between the `SidePane `and `SetupPane`. This could be further extended to the `ReviewPane` in future.
3. All `ModelView` objects selectable in the `WorkflowTree` editor are now `HasTraits` objects. Consequently, they have been renamed `<object>View` rather than `<object>ModelView`.
4. The `MCOParameterView` and `KPISpecificationView` classes now contain a single view to display in the TreeEditor for all MCO parameters and KPIs
5. A new class `ProcessView` now holds a list of `ExecutionLayerViews`, representing all the hierarchical information stored in the list of `ExecutionLayers` for a `Workflow`
6. A new class `CommunicatorView` now holds a list of `NotificationListenerViews`, representing all the `NotificationListeners` for a `Workflow`
7. `NewEntityCreator` has additional options for customising view
8. The Run tool bar icon has been moved to always sit on the far left in both Setup and Review panes
8. Many unit tests have been revised or relocated to better reflect the new structure. A template `TestCase` object is also available to generate a probe `Workflow` object in order to standardise many of the tests.

Known Issues
--------------

- Duplicate error reporting for MCO, this is due to the `WorkflowTree ` containing 3 views for one object (raised in #311)
- KPI validation checks performed in workflow manager rather than BDSS (linked to #307 and force-h2020/force-bdss#213) 

For Future
------------
- Space now available for creating a graph view for `ProcessView` object, as proposed in #282 
- Better use could be made of main view containing with FORCE logo (plugin table could be redesigned depending on #306)
- Syncing of DataSource variable names with their respective KPI / Parameter names (possibly through object that defines the structure and connections in a `ProcessView`)
- Many class methods / unit tests still need docstrings
- Further adoption of `TestCase` for unit tests


force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -0,0 +1,310 @@
+from traits.api import (
```
flongford: This file has been refactored from `kpi_specification_model_view`, though Git recognises it as a new file

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,254 @@
+from traits.api import (
```
flongford: This file has been refactored from `mco_parameter_model_view`, though Git recognises it as a new file

force_wfmanager/ui/setup/communicator/communicator_view.py
```diff
@@ -0,0 +1,65 @@
+from traits.api import (
+    Instance, List, on_trait_change, Event, HasTraits
+)
+from traitsui.api import View, Item
+
+from force_bdss.api import Workflow
+from .notification_listener_view import NotificationListenerView
+
+
+class CommunicatorView(HasTraits):
```
flongford: This is a new class designed to contain the list of `NotificationListener`s in the `Workflow`. At the moment it doesn't add anything to the design, except equalising the hierarchy of objects in the `WorkflowView`: the `MCOView`, `ProcessView` and CommunicatorView` classes all have level 2 status and contain lists of views.

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # ------------------
+    #     View
+    # ------------------
+
+    # The traits_view only displays possible options for
+    # model.name listed in kpi_names. However, it is possible
+    # to directly change model.name without updating kpi_names
+    traits_view = View(
+        UReadonly('model.name'),
```
flongford: This`UReadonly` trait view item used to use an `EnumEditor` to display a list of possible KPI names. However, the new design of the `KPISpecificationView` moves the KPI name selection to before instantiation of the `KPISpecificationModelView`, so the ability to re-select this name feels redundant

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -0,0 +1,308 @@
...
+            return "KPI"
+        return "KPI: {} ({})".format(self.model.name, self.model.objective)
+
+    #: Listeners
+    @on_trait_change('model.[name,type]')
+    def kpi_change(self):
+        self.verify_workflow_event = True
+
+
+class KPISpecificationView(HasTraits):
```
flongford: Considering renaming this `KPIView`

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -0,0 +1,308 @@
...
+        """Listens to variable_names_registry to extract names
+         able to be assigned to kpis"""
+        non_kpi = []
+
+        if self.variable_names_registry is not None:
+            variables_stack = (self.variable_names_registry
+                               .available_variables_stack)
+            for execution_layer in variables_stack:
+                for variable in execution_layer:
+                    kpi_check = variable[0] not in self.kpi_names
```
flongford: Any new KPI must be unique (not listed in the existing `kpi_names`) and only refer to an output variable in the process to optimise (exist in `variable_names_registry.data_source_outputs`). By enforcing this requirement upon creation, it seems more appropriate to remove the ability of the user to change the KPI name after instantiation.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,257 @@
...
+    #: The human readable name of the MCO parameter class
+    label = Property(Unicode(), depends_on="model.[name,type]")
+
+    # ----------
+    #    View
+    # ----------
+
+    def default_traits_view(self):
+        # Obtains traits of BaseMCOParameter that will always be present
+        base_traits = [Item('model.name'),
```
flongford: Currently, we want to always display the `BaseMCOParameter` traits, `name` and `type`. However, these traits are tagged as `visible=False` in their `force_bdss` constructors. Therefore, they will not appear in the default traits view.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,257 @@
...
+                       Item('model.type')]
+
+        # Obtains user-created traits that have been tagged as visible
+        user_traits = [
```
flongford: Typically, user made plugins define MCO parameters with various extra traits that contribute to the MCO optimisation process. If these are tagged as `visible=True` (default) then they will be displayed.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+
+    #: Listeners
+    # Workflow Validation
+    @on_trait_change('model.[name,type]')
+    def parameter_change(self):
+        self.verify_workflow_event = True
+
+
+class MCOParameterView(HasTraits):
```
flongford: New class containing a single view with all `MCOParameterModelView`s

force_wfmanager/ui/setup/mco/mco_view.py
```diff
@@ -0,0 +1,92 @@
...
+    #: Registry of the available variables
+    variable_names_registry = Instance(VariableNamesRegistry)
+
+    # ------------------
+    # Regular Attributes
+    # ------------------
+
+    #: List of MCO parameter and KPI views to be displayed in the TreeEditor
+    #: NOTE: (Has to be a list to be selectable in TreeEditor)
+    mco_options = List(Either(Instance(MCOParameterView),
```
flongford: This list contains single instances of both `MCOParameterView` and `KPISpecificationView`, which are displayed in the `TreeEditor`. They are in the same trait `mco_options`, since this is referenced in the MCO's `TreeNode` as the `children` list trait it displays.

force_wfmanager/ui/setup/new_entity_creator.py
```diff
@@ -70,6 +70,13 @@ class NewEntityCreator(HasStrictTraits):
     #: A message to be displayed if there are no config options
     _no_config_options_msg = ReadOnly(Unicode)
 
+    #: Option for the user to manually control the visibility of the
+    #: configuration option display
+    config_visible = Bool(True)
+
```
flongford: Including new options for controlling the configuration view (used by `MCOParameterView`) and displaying a header to make clear what model options are being displayed

force_wfmanager/ui/setup/process/data_source_view.py
```diff
@@ -39,12 +41,15 @@ class TableRow(HasStrictTraits):
     description = Unicode()
 
     def __init__(self, model, *args, **kwargs):
+        #: FIXME: Child model should not be instantiated before super
```
flongford: Moving the `self.model = model` to after the super instantiation breaks the constructor. I don't know why, but this should be looked at in the future

force_wfmanager/ui/setup/process/process_view.py
```diff
@@ -0,0 +1,100 @@
...
+    HasTraits, Instance, List, on_trait_change, Bool, Event, Unicode
+)
+from force_wfmanager.utils.variable_names_registry import (
+    VariableNamesRegistry
+)
+from .execution_layer_view import ExecutionLayerView
+from force_bdss.api import Workflow
+
+
+class ProcessView(HasTraits):
```
flongford: This new object contains the list of `ExecutionLayers` defined in the `Workflow`. It could also have its own traits view displaying template `Workflows` or a graphical representation of the current `Workflow` structure in the future. It possess the same hierarchy level as the `MCOView`.

force_wfmanager/ui/setup/system_state.py
```diff
@@ -0,0 +1,31 @@
+from traits.api import (
+    HasTraits, Instance, Unicode, Callable
+)
+from force_wfmanager.ui.setup.new_entity_creator import NewEntityCreator
+
+
+class SystemState(HasTraits):
```
flongford: New class acts as a temporary reference to currently selected objects. Not strictly necessary, but abides by the principle of least knowledge guidelines for cross-communicating between tasks / panes

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -206,30 +209,25 @@ def _tool_bars_default(self):
...
         )
 
+    # Overloaded Methods
+
     def create_central_pane(self):
         """ Creates the central pane which contains the layer info part
         (factory selection and new object configuration editors)
         """
-        return SetupPane(workflow_model=self.workflow_model)
+        return SetupPane(
```
flongford: The `SetupPane` now only listens to and displays objects in `SystemState`. It does not hold any information about the `Workflow`

force_wfmanager/ui/setup/mco/tests/test_kpi_specification_view.py
```diff
@@ -0,0 +1,144 @@
...
+    def test_name_change(self):
+
+        self.data_source1.output_slot_info = [OutputSlotInfo(name='T1')]
+        kpi_model_view = self.kpi_view.kpi_model_views[0]
+
+        with self.assertTraitChanges(kpi_model_view, 'label', count=0):
+            self.assertEqual('KPI', kpi_model_view.label)
+
+        with self.assertTraitChanges(kpi_model_view, 'label', count=1):
+            self.workflow.mco.kpis[0].name = 'T1'
```
nicolasap-dm: As commented in the corresponding module, we should define a behaviour for when the output_slot changes in the data source

flongford:
> * I can't seem to delete KPIs.

Try manually selecting a KPI tab first and see if you can delete it - we can try to set a default in which case 

> * In the new entity creator, it used to be possible to double click one of the available ones from the list to use the factory with void configuration. Now it raises an error.

This was a bug, now fixed to display previous behaviour

> * Not sure if it's a bug or if this PR makes the API non-backwards compatible with the extension plugins, but trying to add a "range" parameter to the "Enthought example" MCO raises an error `AttributeError: 'MCOParameterModelView' object has no attribute 'lower_bound'`

This was a bug in the `MCOParameterModelView`, traits view constructor whereby if the `model` trait had any listeners, then they would try to be assigned as a `ModelView` attribute.

In this case, the `RangedMCOParameter` had a `initial_value` trait that depended on another trait `lower_bound`, and the constructor was trying to find `lower_bound` on the `MCOParameterModelView` instead.

It has now been fixed

nicolasap-dm:
> Try manually selecting a KPI tab first and see if you can delete it 

Indeed I can, unless it's the last remaining KPI:

Peek 2019-07-15 14-42
```
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/qt4/button_editor.py", line 121, in update_object
    self.value = self.factory.value
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/editor.py", line 268, in _set_value
    self.ui.do_undoable(self.__set_value, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/ui.py", line 734, in do_undoable
    action(*args, **kw)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/editor.py", line 281, in __set_value
    method(self.ui.info, self.object, name, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/handler.py", line 336, in setattr
    setattr(object, name, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_notifiers.py", line 400, in __call__
    handle_exception(object, trait_name, old, new)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_notifiers.py", line 180, in _handle_exception
    raise excp
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_notifiers.py", line 394, in __call__
    self.handler(*args)
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/ui/setup/mco/kpi_specification_view.py", line 288, in _remove_kpi_button_fired
    self.remove_kpi(self.selected_kpi.model)
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/ui/setup/mco/kpi_specification_view.py", line 311, in remove_kpi
    self.model.kpis.remove(kpi)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 2767, in remove
    list.remove(self, value)
ValueError: list.remove(x): x not in list
```

flongford:
> Indeed I can, unless it's the last remaining KPI:

This should be fixed now, with some additional unit tests to confirm it is behaving as expected

force_wfmanager/ui/setup/mco/kpi_specification_view.py
flongford: Agreed, though this might be best dealt with at the `WorkflowView` or `WorkflowTree` level, since ideally if the user updates a name in a data source this should be picked up by the MCO and the corresponding KPI/Parameter name synced. At the moment, the `KPISecificationView` only checks for trait names in lists, without any knowledge of what the traits actually are.

force_wfmanager/ui/setup/communicator/communicator_view.py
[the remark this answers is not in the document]
nicolasap-dm: > Can we have a docstring?

(not sure if you saw this comment? I might not be familiar with Traits documentation policies, in which case please point me to the right direction)

nicolasap-dm:
Thanks for addressing my comments:

- double clicks now work in the entity creator :heavy_check_mark: 

- removing any KPI works :heavy_check_mark:

- I can add the Range parameter from the Enthought example :heavy_check_mark:
   - however: in this case double clicking raises an error just as before
```
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/qt4/tree_editor.py", line 999, in _on_item_dclicked
    if node.dclick(object) is True:
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traitsui/tree_node.py", line 672, in dclick
    self.on_dclick(object)
  File "/home/ndemitri/repos/force/master/force-wfmanager/force_wfmanager/ui/setup/mco/mco_parameter_view.py", line 268, in add_parameter
    self.model.parameters.append(parameter)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 2673, in append
    raise excp
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 2661, in append
    value = validate(object, self.name, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 1475, in validate
    self.validate_failed(object, name, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 1346, in validate_failed
    self.error(object, name, value)
  File "/home/ndemitri/.edm/envs/force-py36/lib/python3.6/site-packages/traits/trait_handlers.py", line 236, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: Each element of the 'parameters' trait of an ExampleMCOModel instance must be a BaseMCOParameter or None, but a value of <enthought_example.example_mco.parameters.RangedMCOParameterFactory object at 0x7fc29eed6a98> <class 'enthought_example.example_mco.parameters.RangedMCOParameterFactory'> was specified.
```

Unfortunately, the KPI creation still has some issues:
- the new check is not run on existing KPIs. So if I create a KPI, then change a data source output, the interface will never complain (potentially allowing me to launch the optimization in a broken state)
- I don't know if this issue was introduced after my last review: if I have two non-KPI variables, I select the first, and I hit "New KPI", it gets created: good. After that though, other copies of the first (not anymore listed) KPI type will be created regardless of what is selected, and it will be created over and over, if I keep pushing "New KPI". In other words, I can't create one KPI for each of the two listed non-KPI variables.

force_wfmanager/ui/setup/new_entity_creator.py
```diff
@@ -132,40 +142,52 @@ def default_traits_view(self):
...
-                        ),
-                        visible_when="model is not None",
-                        style="custom",
-                        label="Configuration Options",
-                        show_border=True,
-                        springy=True,
+                        UItem("plugins_root",
+                              editor=editor
+                              ),
+                        springy=True
```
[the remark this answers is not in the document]
nicolasap-dm: Sorry, I meant it a bit differently:
- as you have done, I suggested adding the show_border and label

but to work, it also needed
- to be nested into an additional VGroup
- to drop the "UItem("view_header")"
- to drop the html tag in _view_header_defaults

I.e. the first argument to the top=level HSplit would be exactly the content of my message above

force_wfmanager/ui/setup/setup_pane.py
```diff
@@ -33,201 +32,235 @@ class SetupPane(TraitsTaskPane):
...
+                                springy=True
                             ),
+                            # Remove Buttons
+                            UItem(
+                                'remove_entity_btn',
+                                label='Delete Layer',
+                                visible_when=(
+                                    "object.system_state"
+                                    ".selected_factory_name "
+                                    "== 'Data Source'"
```
nicolasap-dm: Can this (and all the remaining long conditions) be made into a property, to complete the refactoring?

flongford: Done

flongford:
I've put in a KPI creator work-around for now by using an `EnumEditor` that only allows the user to select available output variable names for KPIs. 

This replaces the table selection tool and will clear any KPI name that does not match a data source output if that variable is renamed / removed.

An error will also be thrown up if two of the same names are selected, though this will not prevent the user from running the BDSS (likewise if a KPI name is blank). This is something that needs to be addressed in the BDSS verification tests: force-h2020/force-bdss#213

force_wfmanager/ui/setup/new_entity_creator.py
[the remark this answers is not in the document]
flongford: Done, looks better I agree - nice suggestion

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -33,6 +22,10 @@ class KPISpecificationModelView(ModelView):
     #: KPI model
     model = Instance(KPISpecification)
 
+    # Only display name options for existing KPIs (this isn't perfect and
+    # does allow
```
nicolasap-dm: I believe you switched focus mid-sentence :smile:

nicolasap-dm:
I have merged master into this branch: it seems to be working fine, but please have a look!

Note that master uses a different set of dependencies now, so you'll need to reinstall.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,312 @@
...
+    # Dependent Attributes
+    # --------------------
+
+    #: The selected parameter model view
+    selected_parameter = Instance(MCOParameterModelView)
+
+    #: Creates new instances of MCO Parameters
+    parameter_entity_creator = Instance(NewEntityCreator)
+
+    #: Defines if the MCO parameter is valid or not. Updated by
```
jjenthought: Do we need `valid`, `error_message` and `verify_workflow_event` if this class is a grouping of `MCOParameterModelViews`? Hopefully the verification from the parameters themselves should be enough, and if they're okay then the view will be as well.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -0,0 +1,312 @@
...
+        if len(self.parameter_model_views) == 0:
+            self.selected_parameter = None
+
+    # Workflow Validation
+    @on_trait_change('parameter_model_views.verify_workflow_event')
+    def received_verify_request(self):
+        """Pass on request for verify_workflow_event"""
+        self.verify_workflow_event = True
+
+    #: Button actions
```
jjenthought: I think a comment with `#:` gets picked up by the traits specific sphinx extension we use. Though just ignore this if that was intentional!

force_wfmanager/ui/setup/mco/mco_view.py
jjenthought: We should document the form of this attribute, from the two listeners below it looks like it's a 2 element list with `[MCOParameterView, KPISpecificationView]`

force_wfmanager/ui/setup/mco/mco_parameter_view.py
flongford: Ah, good to know - there might be a lot of cleaning up to do in which case

force_wfmanager/ui/setup/mco/mco_parameter_view.py
flongford: Ah, this file contains both `KPISpecificationView` and `KPISpecificationModelView` classes, which is confusing I agree. In #313 this has been inadvertently resolved (along with the MCO parameters).

force_wfmanager/ui/setup/mco/kpi_specification_view.py
flongford: Ah, this file contains both `KPISpecificationView` and `KPISpecificationModelView` classes, which is confusing I agree. In #313 this has been inadvertently resolved (along with the MCO parameters).

force_wfmanager/ui/setup/mco/mco_parameter_view.py
jjenthought: I'd take a look at the documentation output, it might not be getting picked up. Wouldn't worry about fixing it if it isn't.

I know it definitely does for attributes e.g.
```
#: Here's what this trait does
important_trait = Str()
```
but maybe functions aren't included.

force_wfmanager/ui/setup/mco/mco_parameter_view.py
flongford: The traits have been left in at the moment to replace any verification checks referring to a group of KPIs or parameters that are missing in the BDSS (this is more obvious in the `KPISpecificationView`) For instance, do any KPI or parameter share the same name?

Ultimately it would be great to remove these checks from the UI but it feels that this requires some further consideration (hopefully by addressing force-h2020/force-bdss#213

force_wfmanager/ui/setup/mco/kpi_specification_view.py
flongford: Ah, this file contains both `KPISpecificationView` and `KPISpecificationModelView` classes, which is confusing I agree. In #313 this has been inadvertently resolved (along with the MCO parameters).

force_wfmanager/ui/setup/mco/mco_parameter_view.py
flongford: Ok, great - I'll make a new PR with a load of cleaning for sphinx comments anyway

force_wfmanager/ui/setup/mco/kpi_specification_view.py
jjenthought: Haha good to hear! It probably doesn't help that I take so long to review everything!

PR #305, opened 2019-07-12
Better error handling in CI

nicolasap-dm:
This PR affects the `ci` module. It
- replaces the `check_call`s with just `call`s, so that `click` can display a nice error message upon failure

~- backs up the user's `application_memento` file while running test, if it exists, then restores it to its original location~

~Note: I haven't removed the `msg` from here https:// github.com/force-h2020/force-wfmanager/blob/33c7980b25123dfa12d1e38812b607fae4dd9e84/force_wfmanager/tests/test_wfmanager.py#L122-L127 because the test can still fail if `python -m unittest` is run manually, instead of using `ci`.~


jjenthought:
So this might be a bit of a pain since you've gone to the effort here to keep things working, but I think we should drop the application_memento stuff. It seems to cause a fair few issues (some of which might be our fault!), but it doesn't really give us much in return.
Setting `self.always_use_default_layout = True` in wfmanager should let us not have to worry about application_memento stuff

nicolasap-dm:
I won't shead a tear for `application_memento` if it has to go!

If we agree on this change, I'll remove that part of this PR (keeping the `check_call` → `call` changes) and make another PR to suppress `application_memento` and possibly other workarounds that have been created only to deal with it.

jjenthought:
> I won't shead a tear for `application_memento` if it has to go!
Haha yeah me either! Sounds good for the rest of the PR, using the click exceptions is definitely a big improvement

nicolasap-dm:
I have removed the `application_memento` workaround.

nicolasap-dm:
> [I'll] make another PR to suppress application_memento

Note to self: draft in branch [`dont-use-application-memento`](https://github.com/force-h2020/force-wfmanager/tree/dont-use-application-memento)

nicolasap-dm:
Merged master (including APIdoc building in CI, that had conflicts); ready to be merged when approved.

PR #309, opened 2019-07-16
Generate apidoc with `ci docs`, ignore test modules

nicolasap-dm:
This PR
- ~creates a new CI task "`build-apidoc`" which ignores test modules~
- ~pushes the apidocs created accordingly~

See comment below.


jjenthought:
For me, keeping the api doc files in the repository seems likely to cause problems. It means you would want to rebuild and push them with almost every PR, in order to include the new docstrings you write. This sort of thing seems like it would be easy to forget and leave the documentation out of sync with the actual code. This would pass CI though, as the documentation will just build with the old api files.
Distributing the api docs is only (afaik) of use when a user wants to build their own documentation locally (I guess this assumes we have a readthedocs web version available). But if they want to do this, we could just generate current api docs when they run `python -m ci docs` using the new `build-apidoc` command.
tl;dr delet doc/source/api, automate api building as part of `ci docs`

nicolasap-dm:
Changed: now the existing `docs` task takes care of regenerating the apidoc (unless told not to, using `--html-only`). And the apidocs are removed from the repository.

nicolasap-dm:
Absolutely, thanks for spotting that

Issue #307, opened 2019-07-15
Updating referencing variables between Input/Output Slots 


flongford:
Currently, mapping out the input / output slots in the workflow is performed by the `VariableNamesRegistry` and `DataSourceView` objects.

This method could do with a review, considering that it currently does not allow for syncing between variable names in the `DataSourceModel`s and any corresponding `KPI`s or `MCOParameter`s.

Ideally, we would like to create a system where each view is hooked up to any relevant ui objects for name syncing, as well a single object describing this hierarchy for further visualisation.

flongford:
Using the slot objects in `VariableNamesRegistry`, for `available_variables_stack` rather than a tuple of `(name, type)` could allow direct listening from parameters and KPIs to variables

PR #313, opened 2019-07-19
Separate Process from MCO

### LINKED ISSUE #301: Refactor of TreeEditor in WorkflowTree
The current design uses `ModelView` objects as nodes, which is not recommended.

This is likely linked to the issues encountered in #285 and #296.

Refactoring the `TreeEditor` could be done by defining node adaptors for each model class (`DataSource`, `KPI`, `Parameter`...) and their respective model views (http://docs.enthought.com/traitsui/traitsui_user_manual/adapters.html).

In doing so, the `MCO` and `NotificationListener` displays may also be extracted from the workflow tree structure, which seems more suited to display the workflow layers, or a graphical display such as suggested in #282

### LINKED ISSUE #308: Traits instantiated before super class in data_source_view.py
The `TableRow` class in `data_source_view.py`  contains a local `__init__` method that assigns a trait before calling the super class `__init__`. 

    def __init__(self, model, *args, **kwargs):
         self.model = model
         super(TableRow, self).__init__(*args, **kwargs)

Moving the `self.model = model` assignment to after the super instantiation breaks the constructor. I don't know why, but this should be looked at in the future

_Originally posted by @flongford in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0Mjk2NjcyMTgx/pull_requests/unread_timeline_

### LINKED ISSUE #311: Redundant MCO Error handling in TreeEditor
Since #304, using single `TreeEditor` nodes to display all the MCO KPIs and Parameters together means that the `verify_tree` method in `WorkflowTree` needs to be reviewed to suit `KPISpecificationView` and `MCOParameterView` objects.

This originates from the `TreeEditor` being designed to display a hierarchical nest of lists containing objects of the same class. It therefore should be very simple to display the objects in `Workflow` using a class with the same structural hierarchy, `WorkflowView`, that contains objects defining a 'view' for their associated `model=<Workflow object>` traits.

Consequently, the `verify_tree` method goes through each of the objects in `WorkflowView` and calls `model.verify()` to obtain each error message associated with their `model`.

However, the `KPISpecificationView` and `MCOParameterView` classes do not refer to an actual object in the `Workflow` (unlike the `KPISpecificationModelView` and `MCOParameterModelView` classes), so they are given a `model=Instance(BaseMCOModel)` in order to fit the hierarchy.

Because of this, the `verify_tree` method calls `<class: BaseMCOModel>.verify` 3 times and can display redundant error messages.

A straight forward fix could be implemented into the logic of `verfiy_tree`, but ultimately this method (along with the `TreeEditor`) should probably be redesigned to fit the new display system for MCO information.

flongford:
This PR builds on work in #304 to separate the creation of the 'Process' (containing `ExecutionLayer` and `DataSource` classes) from the `BaseMCO` (containing `MCOParameter` and `KPISpecification` classes) in the UI. It therefore closes #301.

It also closes #308 and closes #311 

## Work Done

1. `DataSourceModel` input slot variable names no longer depend on `MCOParameter` names. In doing so, the code causing issue #308 has been removed
2. The `VariableNamesRegistry` has been expanded to return information on both input and output slot names.
3. `MCOParameter` names can only be selected from a list of input slot variables that do not also appear in output slots (ie - they are not generated by the 'Process'). As such the `MCOParameterModelView` object now requires a `_combobox_values` trait, similar to the `KPISpecificationView`.
4. `BaseMCOOptionsView` parent class created for both `MCOParameterView` and `KPISpecificationView` classes
5. `BaseMCOOptionsModelView` parent class created for both `MCOParameterModelView` and `KPISpecificationModelView` classes
6. Verification steps in `WorkflowTree` now deal with KPI and Parameter views independently (fixing #311)
7. (CLN) Dummy class modules have been refactored into separate files inside the `force_wfmanager/tests/dummy_class` directory

## Know Issues

- ~~WfManager `verify_tree` method resets `valid` status for `KPISpecificationView` and `MCOParameterView` to `True` when fired, without re-checking afterwards.~~ 24/7/19 Errors now managed separately
- ~~Errors low down in `Workflow` do not trigger `WorkflowView` to become invalid, which would prevent a BDSS run from being activated.~~ 24/7/19 Child/parent `valid` attributes now updated appropriately
- Due to removal of variable name restrictions, `InputSlotInfo` objects do now not have a handy `Enum` list of available output names to fill, reducing UX

## Future Work

- Allow the user to either enter new `DataSourceModel` variable names, or choose from existing input / output slot values. This could also depend on the structure and order of the Process (input variables for each layer need to have been previously defined as parameters or outputs of the previous layer etc.), and so would be linked to #307


force_wfmanager/ui/setup/mco/mco_parameter_model_view.py
```diff
@@ -0,0 +1,45 @@
+from traits.api import cached_property
+from traitsui.api import (
+    View, Item, InstanceEditor,
+    EnumEditor
+)
+
+from force_wfmanager.ui.setup.mco.base_mco_options_model_view import \
+    BaseMCOOptionsModelView
+
```
nicolasap-dm: :x: blank line

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -1,137 +1,45 @@
...
 
-from force_bdss.api import BaseMCOParameter, BaseMCOModel
-
-from force_wfmanager.ui.ui_utils import get_factory_name
 from force_wfmanager.ui.setup.new_entity_creator import NewEntityCreator
 
+from force_wfmanager.ui.setup.mco.base_mco_options_view import \
+    BaseMCOOptionsView
+from force_wfmanager.ui.setup.mco.mco_parameter_model_view import \
+    MCOParameterModelView
```
nicolasap-dm: same :abc: and blank line nitpick

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -1,137 +1,45 @@
...
-    verify_workflow_event = Event()
-
     # ----------
     # Properties
     # ----------
 
-    #: The human readable name of the MCOParameterView class
-    label = Unicode('MCO Parameters')
+    #: A list names, each representing a variable
+    #: that could become a KPI
```
nicolasap-dm:
```suggestion
    #: that could become a MCO parameter
```

force_wfmanager/ui/setup/mco/mco_parameter_view.py
```diff
@@ -188,14 +92,29 @@ def default_traits_view(self):
...
                 ),
             )
         )
 
         return traits_view
 
+    @cached_property
+    def _get_parameter_name_options(self):
+        """Listens to variable_names_registry to extract
+         possible names for new KPIs"""
```
nicolasap-dm:
```suggestion
         possible names for new MCO parameters"""
```

force_wfmanager/ui/setup/mco/mco_view.py
```diff
@@ -1,13 +1,15 @@
...
-    HasTraits, Either
+    HasTraits
 )
 from traitsui.api import View
 
 from force_bdss.api import BaseMCOModel
 
 from force_wfmanager.utils.variable_names_registry import \
     VariableNamesRegistry
+from force_wfmanager.ui.setup.mco.base_mco_options_view import \
```
nicolasap-dm: please move `.utils` under `.ui`

force_wfmanager/ui/ui_utils.py
```diff
@@ -63,7 +63,7 @@ def class_description(cl, maxlength=80, desc_attribute="description"):
     """ Returns a short description of a class, based on a
     description class attribute (if present) and the class path:
 
-    Examples (maxlenght = 45)::
+    Examples (maxlength = 45)::
```
nicolasap-dm: thanks :bowing_man:

force_wfmanager/tests/test_wfmanager_tasks.py
```diff
@@ -18,7 +18,7 @@
 from .mock_methods import (
     mock_file_reader, mock_file_writer, mock_dialog, mock_return_args
 )
-from .dummy_classes import DummyWfManager
+from force_wfmanager.tests.dummy_classes.dummy_wfmanager import DummyWfManager
```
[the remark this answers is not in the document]
flongford: No worries, good suggestion - there's likely to be some tidying up to do after PyCharm's automatic refactoring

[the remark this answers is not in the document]
jjenthought: For import stuff I've been adding `from` imports after their group i.e.
```
import csv
import subprocess
from collections import defaultdict
```
but i can't remember if this was an Enthough thing or just something I started doing

flongford:
> Issues that are **new** with respect to `master`:
> 
> * The main problem I see with this implementation is that we don't have a type check in place, that we had before: i.e. I can create a parameter using a parameter name from a data source (from the combobox), but then I can make its type the one I like, regardless of that declared in the data source. I might not be aware of the rules for types in the bdss, but could we just populate the type based on the parameter name? Aren't they always going to be related?

Agreed, it also now makes sense to simply get the required input variable and hook it up to the parameter so that both `name` and `type` are synced. I'm currently not sure how big a change that would be, so I'm hesitating to include it in this PR, but there should be some error checking either way.

> * The error "An MCO parameter has no type set" won't go away from the Error box, even if it's resolved

Will look into this, I think it comes from a big in the `verify_tree` method of `WorkflowTree`

nicolasap-dm:
> It think this can be easily fixed in the BDSS, so I will make a separate PR (addressed in force-h2020/force-bdss#226)

With that PR, I confirm that not having a KPI triggers the :warning: sign in the workflow tree.

There are probably still tweaks to do on the verification stage:

- An error (:warning:) on the specific MCO isn't apparently passed back to the parent "MCO" node; this leaves the "Workflow" valid and the "Run" button active.

- An error on either KPI or MCO Parameter triggers a :warning: sign also on the sibling, even if it has no errors (as confirmed by the workflow info box).

Additional cleanup:

- these three constant are defined in several modules but they don't seem to be used
```
_ERROR = "error"
_WARNING = "warning"
_INFO = "information"
```

flongford:
> * An error (⚠️) on the specific MCO isn't apparently passed back to the parent "MCO" node; this leaves the "Workflow" valid and the "Run" button active.

An error on any MCO related object should now be passed back to he `MCOView`, causing an invalid sign (⚠️) to appear. The run is still enabled, however.

> * An error on either KPI or MCO Parameter triggers a ⚠️ sign also on the sibling, even if it has no errors (as confirmed by the workflow info box).

This should now be fixed, and both errors should be dealt with independently.

flongford:
> * An error in the MCO parameters validation (e.g. "An MCO parameter has no type set") doesn't trigger the warning icon, nor it disables the "Run" button.

This bug should now be fixed, it came from using the logic statement:

    if type(<object>) in [<class: parent1>, <class: parent2>:
        ...

rather than 

    if isinstance( <object>, <class: parent1>) or isinstance( <object>, <class: parent2>):
        ...

force_wfmanager/ui/setup/workflow_tree.py
```diff
@@ -746,11 +746,11 @@ def verify_tree(self, errors, start_view=None):
 
             # Further checks for UI objects not relating to an object in the
             # force_bdss Workflow model
-            err_subject_type = type(verifier_error.subject)
 
             # For errors where the subject is an Input/OutputSlotInfo object,
             # check if this is an attribute of the (DataSource) model
-            if err_subject_type in [InputSlotInfo, OutputSlotInfo]:
+            if (isinstance(verifier_error.subject, InputSlotInfo)
```
jjenthought: With `isinstance` you can pass it a tuple of classes like: 
`isinstance(verifier_error.subject, (InputSlotInfo, OutputSlotInfo))`

flongford:
> With just an exception: it is still true that a KPI error causes a ⚠️ sign to appear on both the Parameters and the KPI nodes.

Just noticed this too, the latest commit should have fixed it - though an error message will only appear on the MCO folder node if there are no parameters or KPIs

nicolasap-dm:
Fix confirmed :+1: 

> an error message will only appear on the MCO folder 

This isn't too bad.

PR #317, opened 2019-07-22
Cleanup of Comments for Sphinx Documentation 

flongford:
This PR cleans up comments and import statements affected by #304 

Generally, it removes `#:` tags from non-Sphinx related comments and readers import statements according to Enthought guidelines

    # each group separated by a blank line;
    # alphabetical order within each group and within each line (from X import A, B, C)
    - core library
    - third party (eggs from pypi/eds)
    - ETS (traits, ... envisage)
    - other packages from the current project (here: bdss)
    - this package
    - relative imports (only if written as .relative imports, otherwise they go above)

It's also a chance to raise any other files that might need cleaning.


force_wfmanager/ui/setup/new_entity_creator.py
```diff
@@ -1,3 +1,5 @@
+from envisage.plugin import Plugin
+
```
nicolasap-dm: assuming that my interpretation of the import order rules is correct (thanks for the trust!), this blank line shouldn't be there (both ETS packages)

flongford:
Just pushed a final commit cleaning a load of test files, will merge after CI passed

PR #319, opened 2019-07-26
Handling of plot updates and resizes

### LINKED ISSUE #318: DataView Plot range not updated if X or Y variables are changed
When `x` or `y` are changed on the graph display in the results task, the variable names are updated on the axes but not the range of values for the new data. The user has to therefore manually press the "Reset View" button after changing a plot variable to rescale the axes. 

This suggests that the `reset_view` functionality is working correctly, but the handler that should listen to any changes in `x` or `y` is not working properly.

This bug was originally pointed out in force-h2020/force-bdss-plugin-enthought-example#26

nicolasap-dm:
This PR
- reworks the listeners' logic in `BasePlot`, to close #318, and to remove the `timer.Stop()` call that wasn't working. Now the changes of axes triggers a recenter; the plot ranges are updated based on the timer but only if there were changes in the `evaluation_steps`
- allows switching from colormapped to non-colormapped and back, without losing the pan & zoom

### DIFF
--- a/force_wfmanager/ui/review/plot.py
+++ b/force_wfmanager/ui/review/plot.py
@@ -399,6 +399,27 @@ class BasePlot(BaseDataView):
         self._plot.range2d.y_range.low_setting = y_low
         self._plot.range2d.y_range.high_setting = y_high
 
+    def _get_plot_range(self):
+        """ Helper method to get the size of the current _plot
+
+        Returns
+        ----------
+        x_low: Float
+            Minimum value for x range of plot
+        x_high: Float
+            Maximum value for x range of plot
+        y_low: Float
+            Minimum value for y range of plot
+        y_high: Float
+            Maximum value for y range of plot
+        """
+        return (
+            self._plot.range2d.x_range.low_setting,
+            self._plot.range2d.x_range.high_setting,
+            self._plot.range2d.y_range.low_setting,
+            self._plot.range2d.y_range.high_setting
+        )
+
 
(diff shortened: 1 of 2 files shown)


force_wfmanager/ui/review/plot.py
```diff
@@ -122,11 +119,12 @@ class BasePlot(BaseDataView):
...
+        return Timer(1000, self._check_scheduled_updates)
 
     def __plot_default(self):
         self._plot = self.plot_scatter()
-        self.resize_plot()
+        self.recenter_plot()
+        self.plot_updater.Start()
         return self._plot
 
     def _get_scatter_inspector_overlay(self, scatter_plot):
```
flongford: UX issue: This might be being called when the color plot radio button is being selected, since it removes the axes titles.

The titles will only appear again if either `x` or `y` variables are changed

flongford: This code snippet should refer to `_get_scatter_inspector_overlay` rather than `__plot_default`

force_wfmanager/ui/review/plot.py
```diff
@@ -228,7 +226,7 @@ def update_value_names(self):
             self._plot.x_axis.title = ""
             self._plot.y_axis.title = ""
```
flongford: UX issue: This might be being called when the color plot radio button is being selected, since it removes the axes titles.

The titles will only appear again if either `x` or `y` variables are changed

Issue #219, opened 2018-07-20
VariableNameRegistry depends on creating the data source for type


stefanoborini:
in #201 we are currently forced to instantiate the data source to obtain the slot types and thus associate variable names with their types. This has the side effect of forcing the variable names registry to potentially be fragile if the create_data_source throws exception. 

We need to find an alternative solution to assign the types to variable names, because as it is the design feels fragile, and there's probably something to rationalise.

PR #321, opened 2019-07-30
New Design of VariableNamesRegistry to aid Input/Output slot sync

flongford:
This PR further builds on work in #313 to improve UX and design aspects of the WfManager, working towards a graphical representation of the `Workflow`, as suggested in #282.

It addresses parts of #307 and is linked to #219 

## Work Done

1. `Variable` class defined in `variable_names_registry.py`, which acts as a layer of abstraction between the `DataSource` traits and those chosen as KPIs and parameters
2. Attribute `VariableNameRegistry.variable_registry` is a dictionary containing `Variable` objects that refer to input/output slots on the `DataSources`. Consequently, the names of each variable have persistence and can be tracked if they are changed. Additionally, it allows for the same input variable to be referenced multiple times in the `DataSource` slots, but dealt with as a single entity in the MCO.
3. Traits on `KPISpecification` and `MCOParameter` objects should now be automatically updated if these are changed on the corresponding `DataSources`
4. `MCOOptionsModelView` classes now contain `selected_variable` and `available_variable` traits, rather than `_combobox_values`. Consequently, their views now use `InstanceEditors` rather than `EnumEditors`.

## Know Issues
- Error message handling from `Variable` class needs to be hooked up to `WorkflowTree`
- `VariableNamesRegistry` has redundant methods that could be removed / replaced
## Future Work

- (Ongoing from #313) Allow the user to either enter new `DataSourceModel` variable names, or choose from existing input / output slot values.

### DIFF
--- a/force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
+++ b/force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
@@ -9,7 +9,9 @@ from traitsui.api import (
 from force_bdss.api import (
     KPISpecification, BaseMCOParameter, Identifier
 )
-from force_bdss.local_traits import CUBAType
+from force_wfmanager.utils.variable_names_registry import (
+    Variable
+)
 
 
 class BaseMCOOptionsModelView(ModelView):
@@ -22,11 +24,11 @@ class BaseMCOOptionsModelView(ModelView):
     model = Either(Instance(KPISpecification),
                    Instance(BaseMCOParameter))
 
-    #: Only display name and type options for available variables
-    # FIXME: this isn't an ideal method, since it requires further
-    # work arounds for the name validation. Putting better error
-    # handling into the force_bdss could resolve this.
-    available_variables = List(Tuple(Identifier, CUBAType))
+    # Variable selected by the UI to hook model up to
+    selected_variable = Instance(Variable)
+
+    #: List of available variables for UI selection
+    available_variables = List(Variable)
 
     # ------------------
     # Regular Attributes
@@ -50,10 +52,6 @@ class BaseMCOOptionsModelView(ModelView):
     #     Properties
     # ------------------
 
-    #: Values for model.name EnumEditor in traits_view
-    _combobox_values = Property(List(Identifier),
-                                depends_on='available_variables')
-
     def __init__(self, model=None, *args, **kwargs):
         super(BaseMCOOptionsModelView, self).__init__(*args, **kwargs)
         if model is not None:
@@ -63,27 +61,8 @@ class BaseMCOOptionsModelView(ModelView):
     #     Listeners
     # ------------------
 
-    def _get__combobox_values(self):
-        """Update combobox_values based on available variable names"""
-        _combobox_values = []
-        if self.available_variables is not None:
-            for variable in self.available_variables:
-                _combobox_values.append(variable[0])
-
-        return _combobox_values
-
     # Assign an on_trait_change decorator to specify traits to listen to
     # in child class implementation
     def model_change(self):
         """Raise verify workflow event upon change in model"""
         self.verify_workflow_event = True
-
-    @on_trait_change('model.name,_combobox_values')
-    def _check_model_name(self):
-        """Check the model name against a compiled list of available
-         output variable names. Clear the model name if a matching
-         output is not found"""
-        if self.model is not None:
-            if self._combobox_values is not None:
-                if self.model.name not in self._combobox_values + ['']:
-                    self.model.name = ''
--- a/force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
+++ b/force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
@@ -1,13 +1,12 @@
 from traits.api import (
-    Bool, Event, Instance, List, Property, Unicode,
-    on_trait_change, Either, Tuple
+    Bool, Event, Instance, List, Unicode, Either, on_trait_change
 )
 from traitsui.api import (
     ModelView
 )
 
 from force_bdss.api import (
-    KPISpecification, BaseMCOParameter, Identifier
+    KPISpecification, BaseMCOParameter
 )
 from force_wfmanager.utils.variable_names_registry import (
     Variable
@@ -60,6 +59,21 @@ class BaseMCOOptionsModelView(ModelView):
     # ------------------
     #     Listeners
     # ------------------
+    @on_trait_change('selected_variable,available_variables')
+    def _check_available_variables(self):
+
+        if self.selected_variable is not None:
+            if self.selected_variable not in self.available_variables:
+                self.selected_variable = None
+
+    @on_trait_change('selected_variable')
+    def _check_selected_model(self):
+
+        if self.selected_variable is None:
+            if self.model is not None:
+                self.model.name = ''
+                if isinstance(self.model, BaseMCOParameter):
+                    self.model.type = ''
 
     # Assign an on_trait_change decorator to specify traits to listen to
     # in child class implementation
--- a/force_wfmanager/ui/setup/mco/kpi_specification_model_view.py
+++ b/force_wfmanager/ui/setup/mco/kpi_specification_model_view.py
@@ -2,7 +2,7 @@ from traits.api import (
     Property, Instance, Unicode, on_trait_change
 )
 from traitsui.api import (
-    Item, View, EnumEditor
+    Item, View, InstanceEditor
 )
 
 from force_wfmanager.ui.setup.mco.base_mco_options_model_view import \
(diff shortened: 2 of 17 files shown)


force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
```diff
@@ -22,11 +23,11 @@ class BaseMCOOptionsModelView(ModelView):
     model = Either(Instance(KPISpecification),
                    Instance(BaseMCOParameter))
 
-    #: Only display name and type options for available variables
-    # FIXME: this isn't an ideal method, since it requires further
-    # work arounds for the name validation. Putting better error
-    # handling into the force_bdss could resolve this.
-    available_variables = List(Tuple(Identifier, CUBAType))
+    # Variable selected by the UI to hook model up to
```
flongford: Here we introduce the `Variable` class as selectable objects for MCO parameters or KPIs. These objects are generated in the `VariableNamesRegistry` and updated each time an input of output slot is edited.

force_wfmanager/ui/setup/mco/base_mco_options_model_view.py
```diff
@@ -62,28 +59,24 @@ def __init__(self, model=None, *args, **kwargs):
...
+                self.selected_variable = None
 
-        return _combobox_values
+    @on_trait_change('selected_variable')
+    def _check_selected_model(self):
+
+        if self.selected_variable is None:
+            if self.model is not None:
+                self.model.name = ''
+                if isinstance(self.model, BaseMCOParameter):
```
flongford: Instances of type `KPISpecification` do not have a `type` trait

force_wfmanager/ui/setup/mco/kpi_specification_model_view.py
```diff
@@ -26,8 +26,12 @@ class KPISpecificationModelView(BaseMCOOptionsModelView):
     # model.name listed in kpi_names. However, it is possible
     # to directly change model.name without updating kpi_names
     traits_view = View(
-        Item('name', object='model',
-             editor=EnumEditor(name='object._combobox_values')),
+        Item('selected_variable',
```
flongford: Rather than using an `EnumEditor`, we can select between `Variable` instances from the list `available_variables`

force_wfmanager/ui/setup/mco/kpi_specification_model_view.py
```diff
@@ -40,9 +44,17 @@ def _get_label(self):
...
+            self.model.name, self.model.objective
+        )
 
     #: Listeners
-    @on_trait_change('model.[name,objective]')
-    def kpi_model_change(self):
-        self.model_change()
+    @on_trait_change('model.name,'
+                     'selected_variable.name')
+    def selected_variable_change(self):
```
flongford: This is an extra check, since the user should not be able to edit the `model.name` through the view anyway

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -100,44 +96,35 @@ def _model_views_default(self):
...
-                )
-                for kpi in self.model.kpis
-            ]
+            # Add all MCO KPIs as ModelViews
+            for kpi in self.model.kpis:
+                model_view = KPISpecificationModelView(
+                        model=kpi,
+                        available_variables=self.kpi_name_options
+                    )
+                for var in self.kpi_name_options:
```
flongford: During instantiation, if KPI already has a name, look for a matching `Variable` object to point to

force_wfmanager/ui/setup/mco/kpi_specification_view.py
```diff
@@ -100,44 +96,35 @@ def _model_views_default(self):
...
-            outputs = self.variable_names_registry.data_source_outputs
-            inputs = self.variable_names_registry.data_source_inputs
-
-            kpi_name_options += (
-                [output_ for output_ in outputs
-                 if output_ not in inputs]
-            )
+            variables = self.variable_names_registry.available_variables
+            for variable in variables:
+                if len(variable.input_slots) == 0:
```
flongford: Here we define our criteria for choosing a KPI - in this case no intermediate `Variables` are allowed (those consumed in the execution layer process)

force_wfmanager/ui/setup/process/data_source_view.py
```diff
@@ -42,56 +39,28 @@ class TableRow(HasStrictTraits):
...
+    model = Instance(InputSlotInfo)
 
-    @on_trait_change('name')
-    def update_model(self):
-        """Synchronises the model if the user changes a model object's name
-        via the UI
-        """
-        self.model.input_slot_info[self.index].name = self.name
+    # Available names for input variables
+    available_names = List(Identifier)
```
flongford: At the moment this is unused, but could be useful in the future if allowing the user to select between existing variable names

force_wfmanager/ui/setup/process/data_source_view.py
```diff
@@ -42,56 +39,28 @@ class TableRow(HasStrictTraits):
...
-    #     Listeners
+    # Required Attributes
     # -------------------
 
-    @on_trait_change('model.input_slot_info[]')
-    def update_view(self):
-        """Synchronises the InputSlotRow with the underlying model"""
-        self.name = self.model.input_slot_info[self.index].name
+    #: Model of the evaluator
+    model = Instance(InputSlotInfo)
```
flongford: The slot row names are now pointing directly to their `model.name`

force_wfmanager/ui/setup/workflow_tree.py
```diff
@@ -429,6 +429,9 @@ def perform_verify_workflow_event(self):
                 self.workflow_view.mco_view[0]
                 .kpi_view.verify_model_names()
             )
+
+        errors += self.workflow_view.variable_names_registry.verify()
```
flongford: Obtain errors referring to input and output slots that are raised by generating `Variable` objects

force_wfmanager/utils/variable_names_registry.py
```diff
@@ -2,221 +2,326 @@
...
-    #: representing the input variables produced by that execution layer
-    data_source_input = List(Tuple(Identifier, CUBAType))
+    # Layer index at which variable is generated
+    layer = Int
 
-    #: For each execution layer, there will be a list of (name, type) pairs
-    #: representing the output variables produced by that execution layer
-    data_source_output = List(Tuple(Identifier, CUBAType))
+    # DataSource that contains Variable as an output
+    origin = Instance(BaseDataSourceModel)
```
flongford: At the moment this trait is unused, but would be handy for creating a graphical representation of the execution layer process, since we have a reference of which `DataSources` are hooked up to each other

force_wfmanager/utils/variable_names_registry.py
```diff
@@ -2,221 +2,326 @@
...
 
     def __init__(self, workflow, *args, **kwargs):
         super(VariableNamesRegistry, self).__init__(*args, **kwargs)
         self.workflow = workflow
 
+    def __variable_registry_default(self):
+        """Default value for _variable_registry dictionary: split into two inner
+        dictionaries, holding defined (created by a DataSource output slot
+        and undefined (requiring to be created by a MCOParameter) Variables"""
+        return {'defined': {},
```
flongford: I don't know if these are the most descriptive names (`defined` and `undefined`), so feel free to suggest alternatives

force_wfmanager/utils/variable_names_registry.py
```diff
@@ -2,221 +2,326 @@
...
+    # ------------------
+
+    label = Property(Unicode, depends_on='output_slot.name,type')
+
+    # ------------------
+    #     Listeners
+    # ------------------
+
+    def _get_label(self):
+        return f'{self.type} {self.name}'
```
nicolasap-dm: Discussed offline: suggest using a tuple (type, name) to avoid ambiguity

flongford: Included in latest commit

flongford:
> * opening a saved workflow which has a DataSource that has a variable appearing both as input and as output, will trigger errors in the checks, but in `master` it would run

This is a general design concern - previously using redundant name/type combinations did not raise an error, but using `Variable` objects to store and connect together `DataSources` in the UI requires some way of identifying these values uniquely.

> * in the Data Source, changing the type of one variable will wipe entirely all variables and won't allow re-entering them (I need to close the wfmanager or to remove the DataSource)

Should now be fixed in latest commit, was a bug in `data_source_view.py`

flongford:
>I have a concern about the user not being able to anticipate whether a rename will have effect on other variables or not. I'm wondering whether a text box should appear (as a nudge) when initiating a edit in the DataSource tables' cells, which reads:
>- if editing an Input variable: either "`This input variable is hooked to an Output from Layer \<N\>. Renaming will remove the hook`" or "`This input variable is currently unhooked`" (not necessarily), or even "`The MCO parameter linked to this input will be renamed as well`"
>- if editing an Output variable: either "`[\<N\> input variables] [and] [a KPI] are hooked to this output variable; they will be renamed too.`", with an option to "[ ] `unhook inputs`", or nothing

A new "status" box in the `TableEditor` could be included with these messages. 

However, doing so would require an additional way to communicate between the `Variable` objects and the input and output slot views, rather than the model itself. In truth, this would allow for easier listening to the UI in general, though the work required would probably be beyond this PR

PR #324, opened 2019-08-02
Custom UIs

jjenthought:
~This is mostly functional but still needs tests!~ We now have tests!
The PR:
- Adds a new `UIExtensionPlugin` type which is basically the `BaseExtensionPlugin` + GUI stuff
- Adds the `ContributedUI` class, which is where the user defines their workflow data and the UI representation for it. It also contains some functionality for checking if the plugins required by this UI are actually available (and makes it unselectable if not) NOTE: At the minute this PR simply doesn't display the UI as an option, it may be nicer to display it but greyed out or with a helpful error message
- Adds a modal dialogue for choosing the UI, accessible from the main toolbar in `Setup`


flongford:
Nice! I like this enhancement a lot.

@nicolasap-dm Maybe we could also use a similar setup to the custom data views if we need to explore a different way to contribute these?

force_wfmanager/ui/contributed_ui/contributed_ui.py
```diff
@@ -0,0 +1,134 @@
...
+    search_term: Any
+        Anything that can be used as a dictionary key
+    results: List, optional
+        A list storing the current results. Used when search is called
+        recursively e.g. for a dict which has dicts as values.
+    """
+    if not results:
+        # Initial empty results list
+        results = []
+    if isinstance(input, dict):
```
jjenthought: I feel like there has to be a better way of doing this..

force_wfmanager/ui/contributed_ui/ui_select_modal.py
```diff
@@ -0,0 +1,146 @@
...
+    """A Handler to provide correct closing behaviour for UISelectModal.
+    If the OK button is pressed, open a window with the view for that UI.
+    """
+    def close(self, info, is_ok):
+        if not is_ok:
+            info.object.selected_ui = None
+        return True
+
+
+class UISelectModal(HasRequiredTraits):
```
flongford: Quick terminology question, is this a "modal" or a "model"?

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -213,6 +221,13 @@ def _tool_bars_default(self):
                     method="open_plugins",
                     image_size=(64, 64)
                 ),
+                TaskAction(
```
flongford: I suppose later on we could include an initialisation window at the beginning of the app to just run any custom UI outside of `WfManagerSetupTask`? Ultimately it just creates a workflow file that can be ran by the headless `force-bdss` and then analysed on `WfManagerReviewTask`.

This could allow us to hide certain complexities from the user, depending on their technical knowledge.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -666,3 +681,39 @@ def switch_task(self):
 
     def exit(self):
         self.window.close()
+
+    # Custom UI Methods
+    def ui_select(self):
```
flongford: We could also move `selected_contributed_ui` to become an attribute of `SystemState` if we wanted to separate the custom UI from the setup pane. This isn't necessary at the moment, but could be done at a later stage depending on how we choose to deliver a 'streamline' version of the app.

force_wfmanager/wfmanager_setup_task.py
flongford: Additionally, running the custom UI inside `force_wfmanager` means including `force_wfmanager` as a requirement for any plugin that contributes a custom UI (force-h2020/force-bdss-plugin-enthought-example#28). Therefore it would be even better to allow the custom UI to run as a stand alone app, as well as within the `force_wfmanager`. However, this can be achieved in later PRs.

PR #329, opened 2019-09-26
ENH: Refactor DataSourceView to reflect changes in force-bbss

flongford:
The PR requires force-h2020/force-bdss/pull/230

Since the synchronization between slot attributes on `BaseDataSource` and `BaseDataSourceModel` classes is now performed in the `force-bdss`, this PR removes unnecessary routines in `DataSourceView` that performed the same task.

Consequently, all attributes of a variable are contained in either a `InputSlotInfo` or `OutputSlotInfo` class, and so these traits are referenced directly in any `TableRow` TraitsUI handler that displays them.

Some unit tests that were breaking due to force-h2020/force-bdss/pull/230 have also been tidied up. Most notably, since each `BaseDataSourceModel` now instantiates an associated `BaseDataSource` upon initiation, it is no longer safe to mix subclasses of these. For example, a `DummyDataSourceModel` should not be instantiated using a `ProbeDataSourceFactory` that contains a reference to a `ProbeDataSource`, as it not guaranteed that the model class has the correct attributes to successfully call `probe_data_source.slots(dummy_model)`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
     - if [[ ${TRAVIS_OS_NAME} == "linux" ]] ; then export PATH=${HOME}/edm/bin:${PATH} ; fi
     - if [[ ${TRAVIS_OS_NAME} == "osx" ]]; then export PATH="${PATH}:/usr/local/bin" ; fi
     - edm install --version 3.6 -y click setuptools
-    - git clone git://github.com/force-h2020/force-bdss.git
+    - git clone git://github.com/force-h2020/force-bdss.git -branch enh/refactor_slots
     - pushd force-bdss && edm run -- python -m ci build-env && edm run -- python -m ci install && popd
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -27,7 +27,7 @@ before_install:
     - if [[ ${TRAVIS_OS_NAME} == "linux" ]] ; then export PATH=${HOME}/edm/bin:${PATH} ; fi
     - if [[ ${TRAVIS_OS_NAME} == "osx" ]]; then export PATH="${PATH}:/usr/local/bin" ; fi
     - edm install --version 3.6 -y click setuptools
-    - git clone git://github.com/force-h2020/force-bdss.git
+    - git clone git://github.com/force-h2020/force-bdss.git --branch enh/refactor_slots
```
dpinte: just one comment! we likely want to get this removed before merging to master!

PR #331, opened 2019-10-11
Updates to BDSS BaseFactory Refactor

flongford:
This PR implements updates required by the refactoring of `force_bdss` in force-h2020/force-bdss/pull/232

### Change Log
- `NewEntityCreator` class `_plugins_root_default` method refactored
- All `factory.plugin.id` references updated to `factory.plugin_id`
- All `factory.plugin.name` references updated to `factory.plugin_name`

**Note:** this branch is pinned to `force-bdss/enh/refactor-factory-plugin` and should not be merged before force-h2020/force-bdss/pull/232 is closed

PR #332, opened 2019-11-01
FIX: Removed broken factory.plugin.id reference

flongford:
A broken reference to `factory.plugin.id` has been removed from the `VariableNamesRegistry.update_available_variables_stacks` method. This was previously missed in #331 as it only occurred during an exception that was not covered in the test suite.

It is also not advisable for a change handler to raise an exception, so creating a unit test for this case is not as beneficial as refactoring the code (taken care of in #329)

### DIFF
--- a/force_wfmanager/utils/variable_names_registry.py
+++ b/force_wfmanager/utils/variable_names_registry.py
@@ -230,7 +230,7 @@ class VariableNamesRegistry(HasStrictTraits):
                         "in plugin '{}'. This may indicate a programming "
                         "error in the plugin".format(
                             data_source_model.factory.id,
-                            data_source_model.factory.plugin.id))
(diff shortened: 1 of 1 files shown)


Corwinpro:
I can see some other appearances of the `plugin_id` variables in the code. I am sure these are not related to this PR, but could you please confirm we don't have to remove the rest of them as well?

flongford:
> I can see some other appearances of variables named `plugin.id` in the code. I am sure these are not related to this PR and refer to the actual `id` of a `plugin`, but could you please confirm we don't have to remove the rest of them as well?

Yes, that's correct - the references in `plugin_dialog.py` are safe as they directly refer to the `id` of a `Plugin` instance, rather than `factory.plugin.id`/

Issue #333, opened 2019-12-19
MCOProgressEvent should accept metadata instead of weights


Corwinpro:
The signature of `MCOProgressEvent` is [hard-coded](https://github.com/force-h2020/force-wfmanager/blob/f9ca5b47189b1a23dff3563716e11f0bed345693/force_wfmanager/wfmanager_setup_task.py#L470) into the workflow manager, and therefore it is impossible to change the event behaviour solely on the `force-bdss` level. In particular, whenever this event occurs, the WFManager expects the event to have the `weights` attribute, which may be absent for various reasons.

We can consider two options:
- Refactor `_server_event_mainthread` to parse metadata instead of just weights. This expands the range of possible data we can pass to the workflow manager, but it seems wrong to make `wfmanager_setup_task.py` responsible for this.
-  Implement a serialisation method for `MCOStartEvent`, `MCOProgressEvent` in a way the classes themselves are responsible for the data representation

Corwinpro:
This is also related to [this issue](https://github.com/force-h2020/force-bdss/issues/239)

Corwinpro:
[This BDSS PR](https://github.com/force-h2020/force-bdss/pull/244) introduced major changes to `MCOEvents` initialization. Now it should be easier to approach this issue.

flongford:
I believe this can be closed @Corwinpro? I can create a separate issue for moving the `BaseDriverEvent` serialization and deserialization code onto the classes themselves

PR #339, opened 2020-01-16
DEV: MCOEvent data display in the analysis pane

Corwinpro:
This PR approaches #333 

This PR expects [this issue](https://github.com/force-h2020/force-bdss/issues/247) to be resolved first.

### Summary

As mentioned in the issue #333 , the `_server_event_mainthread` method of the `WfManagerSetupTask` setup task was processing the data coming from the `MCOEvent`s. That was an undesired behaviour, since the `MCOEvent`s may have different signature (some plugins don't use weights as attribute of `MCOEvent`s, or use something different). Since the `serialize` method was introduced to `MCOEvent` in [this PR](https://github.com/force-h2020/force-bdss/pull/244), now it is very easy to use that internal method instead of parsing event data inplace.

### Done

- Now `_server_event_mainthread` uses `MCOStartEvent.serialize()` and `MCOProgressEvent.serialize()` methods and doesn't process the internal events data itself
- Tests updated

#### In Force-BDSS
- [X] Remove the redundant `weights` attribute in the `MCOProgressEvent` class
- [X] Implement a child class `WeightedMCOProgressEvent(MCOProgressEvent )` that includes `weights` as an argument
- [X] Change the `notify_new_point` method such that it creates an appropriate progress event, instead of a base `MCOProgressEvent`

#### In plugins
- [X] Update `MCO` classes and the events they generate when a new point is generated


Corwinpro:
Note: the standard `MCOProgressEvent` does not generate the `weights` anymore, and therefore the weights would not be displayed in the table.

flongford:
There's some fixes to implement in the `enthought-example-plugin` and `itwm-example-plugin`, I've made a PR for one https://github.com/force-h2020/force-bdss-plugin-enthought-example/pull/39 and am working on the other

flongford:
> There's some fixes to implement in the `enthought-example-plugin` and `itwm-example-plugin`, I've made a PR for one [force-h2020/force-bdss-plugin-enthought-example#39](https://github.com/force-h2020/force-bdss-plugin-enthought-example/pull/39) and am working on the other

ITWM case updated https://github.com/force-h2020/force-bdss-plugin-itwm-example/pull/70

PR #340, opened 2020-01-21
FIX: New Workflow serialization / deserialization methods

flongford:
This PR applies required changes to `Workflow` class saving / loading routines that are introduced in https://github.com/force-h2020/force-bdss/pull/266

It should therefore not be merged before https://github.com/force-h2020/force-bdss/pull/263

Main changes include:

- Removal of redundant `open` function calls in `workflow_io` routines, now files are read and parsed within `WorkflowWriter` and `WorkflowReader` classes
- Use of `Workflow.from_json` method to serilalize and deserilalize a `Workflow` instance in `WfManagerReviewTask` class, rather than writing and reading to file
- Replace the deprecated `WorkflowReader.from_dict` method with `WorkflowReader.parse_data` and `Workflow.from_json` methods `ContributedUI.workflow_data`
- Simplification of many unit tests affected by this change (removal of `mock_open`)


Corwinpro:
The test `tests/test_wfmanager_setup_task.py:test_save_workflow` is failing because the mock object  defined by the line 40 `WORKFLOW_WRITER_PATH = 'force_wfmanager.io.workflow_io.WorkflowWriter'` is not used anymore.

Corwinpro:
@flongford I noticed you added explicit instantiation of a `Workflow` object in `load_project_file` instead of the `WorkflowReader.read`. Could we please discuss this? Is it better than delegating this process to the `WorkflowReader.read`? I am sure you have ideas about that - I can only think of an easier testing side of it.

flongford:
> @flongford I noticed you added two lines with explicit instantiation of a `Workflow` object in `load_project_file` instead of the `WorkflowReader.read`. Could we please discuss this? Is it better than delegating this process to the `WorkflowReader.read`? I am sure you have ideas about that - I can only think of an easier testing.

Mainly just so that we do not read the same file twice. However, I can also imagine that if the "project file" becomes more complicated than just a simple JSON then there is no guarantee that it can be opened by `WorkflowReader.read` in the future (although it would still contain data that could be parsed by `WorkflowReader.parse_data`).

Corwinpro:
> Mainly just so that we do not read the same file twice. However, I can also imagine that if the "project file" becomes more complicated than just a simple JSON then there is no guarantee that it can be opened by `WorkflowReader.read` in the future (although it would still contain data that could be parsed by `WorkflowReader.parse_data`).

Very good point on reading the same file twice. I guess I miss something, but don't we read the file only once when using `workflow_model = reader.read(file_path)`?

Also, I would image that `reader = WorkflowReader(factory_registry)` should be then replaced with something like `reader = WorkflowFile.reader` (or similar), such that the when the "project file" becomes more complicated, it should be the `WorkflowFile`'s responsibility to handle that.

flongford:
> Very good point on reading the same file twice. I guess I miss something, but don't we read the file only once when using `workflow_model = reader.read(file_path)`?

I believe the same JSON `file_path` file is being read in line 58 of `project_io` to obtain the `analysis_model_dict` data.

> Also, I would image that `reader = WorkflowReader(factory_registry)` should be then replaced with something like `reader = WorkflowFile.reader` (or similar), such that the when the "project file" becomes more complicated, it should be the `WorkflowFile`'s responsibility to handle that.

Yes that would be better, and avoid having to create multiple instances of `WorkflowReader`. I will think about how to fit in the `WorkflowFile` class.

Corwinpro:
> > Very good point on reading the same file twice. I guess I miss something, but don't we read the file only once when using `workflow_model = reader.read(file_path)`?
> 
> I believe the same JSON `file_path` file is being read in line 58 of `project_io` to obtain the `analysis_model_dict` data.
> 
> > Also, I would image that `reader = WorkflowReader(factory_registry)` should be then replaced with something like `reader = WorkflowFile.reader` (or similar), such that the when the "project file" becomes more complicated, it should be the `WorkflowFile`'s responsibility to handle that.
> 
> I believe the same JSON file_path file is being read in line 58 of project_io to obtain the analysis_model_dict data.

That is true, sorry I didn't spot it. Perhaps an advanced WorkflowReader/Writer classes can be added to the workflow manager, that would handle this extra piece of data.

flongford:
> > I believe the same JSON `file_path` file is being read in line 58 of `project_io` to obtain the `analysis_model_dict` data.
> 
> That is true, sorry I didn't spot it. Perhaps an advanced WorkflowReader/Writer classes can be added to the workflow manager, that would handle this extra piece of data.

Yes, I think in future we should have both `ProjectReader` and `ProjectWriter` classes implemented, though maybe they are a bit outside the current scope of this PR

Corwinpro:
LGTM. Happy to merge both the BDSS and WfManager PRs when you feel it is all ready to go.

.travis.yml
```diff
@@ -32,6 +32,7 @@ before_install:
     - edm install --version 3.6 -y click setuptools
     - git clone git://github.com/force-h2020/force-bdss.git
     - pushd force-bdss
+    - git checkout dev/workflow-serialization-264
```
flongford: Git checkout reminder

force_wfmanager/io/project_io.py
```diff
@@ -57,10 +57,12 @@ def load_project_file(factory_registry, file_path):
...
+    analysis_model_dict = project_json['analysis_model']
 
-        reader = WorkflowReader(factory_registry)
-        fp.seek(0)
-        workflow_model = reader.read(fp)
+    reader = WorkflowReader(factory_registry)
+    workflow_data = reader.parse_data(project_json)
+
+    workflow_model = Workflow.from_json(
+        factory_registry, workflow_data)
```
Corwinpro: (STY: this can fit on one line)

force_wfmanager/tests/test_wfmanager_review_task.py
```diff
@@ -59,74 +69,100 @@ def test_save_project_failure(self):
...
-            wf_manager=DummyWfManagerWithPlugins())
+            wf_manager=DummyWfManagerWithPlugins()
+        )
 
     def test_discover_data_views(self):
         # One default plot type plus three contributed
         self.assertEqual(
-            len(self.review_task.central_pane.available_data_views), 4)
+            len(self.review_task.central_pane.available_data_views), 4
+        )
```
Corwinpro: (TST: should this be `4, len(self.review_task.central_pane.available_data_views)` instead?)

PR #346, opened 2020-01-30
DEV: Reporting general type data values

Corwinpro:
### Summary

We approach to allow the progress events to contain non numerical data values (e.g., strings), such that the numerical data can be still displayed in the plot section. Now, passing `str` or any other non-`int, float` data to `DataValue(value=...)` should not break the display section of the WorkflowManager.

### Done
- Added `numerical_value_names` attribute to the `AnalysisModel`. This attributes represents the columns that contain only `Either(int, float)` data, and therefore can be plotted.
- The `_value_names` of the `Plot` class delegates to the `AnalysisModel.numerical_value_names`. The `Plot` allows only the `_value_names` columns with numerical data to be chosen as axis. 

### Notes
- The `numerical_value_names` attribute is populated **after** the first `MCOProgressEvent`, when the first data row is available, i.e. after the `add_evaluation_step` is called. This means, that no axis of the `Plot` can be chosen in between the `MCOStartEvent` and `MCOProgressEvent` notifications.


force_wfmanager/model/analysis_model.py
```diff
@@ -111,6 +113,27 @@ def add_evaluation_step(self, evaluation_step):
         self._evaluation_steps.append(evaluation_step)
         self._export_enabled = True
 
+    @on_trait_change("_evaluation_steps[]")
+    def update_numerical_value_names(self):
+        if len(self._evaluation_steps) == 0:
+            return
```
flongford:
```suggestion
            return []
```

Corwinpro:
@flongford The main bug was that having `@on_trait_change('some_attribute')` which listens to a `some_attribute = DelegatesTo('some_object')` attribute is not equivalent to `@on_trait_change('some_object')`.

Should be fixed now.

force_wfmanager/ui/review/tests/test_plot.py
```diff
@@ -274,54 +287,49 @@ def test_recenter_plot(self):
...
-class TestBasePlot(TestAnyPlot, unittest.TestCase):
-
-    def setUp(self):
-        super(TestBasePlot, self).setUp()
-        self.plot = BasePlot(analysis_model=self.analysis_model)
-
-
-class TestPlot(TestAnyPlot, unittest.TestCase):
-
+class TestPlot(TestAnyPlot):
```
flongford: Does this mean that the unit tests in the `TestAnyPlot` class will be performed twice?

Corwinpro: I believe so, but the `setUp` method generates different `Plot` subclass.

PR #351, opened 2020-02-07
MAINT: Plot axis view

Corwinpro:
### Summary

This PR refactors the `BasePlot` class. The aim was to make implementation of the axis and data updates more transparent, and eliminate multiple calls of traits listeners all over the code.

We also refactor the way the view is generated, which allows better and easier customization.

#### View
- Added `default_traits_view` method. It generates the `VGroup` as before, but the first entry of the group is not provided by the `self.axis_hgroup` attribute. If we want to create a different view for the axis choice panel, we can implement a custom `self.axis_hgroup`.  For example, the `BasePlot` has:
```
    def _axis_hgroup_default(self):
        return HGroup(
            Item("x", editor=EnumEditor(name="displayable_value_names")),
            Item("y", editor=EnumEditor(name="displayable_value_names")),
        )
```
The `Plot` class overwrites the axis panel, adding a `"color_option"` `UItem`:
```
    def _axis_hgroup_default(self):
        return HGroup(
            Item("x", editor=EnumEditor(name="displayable_value_names")),
            Item("y", editor=EnumEditor(name="displayable_value_names")),
            UItem("color_options"),
        )
```
- The `x` and `y` attributes are not `Enum` anymore, but plane strings with `EnumEditor` on the frontend. The backend is now responsible for updating the axis properly.

#### General refactoring
- The `update_required` attribute is now `True` by default, which allows the plot to be updated properly if the simulation was started from the `Setup` section.
- Axis update refactored: now we have separate methods which are responsible for the `x` and `y` axis update. Previously one method was handling that. This means, that when a new `x` axis is chosen, the `y` axis bounds are not affected.
- Plot is not recentered every second anymore. This should be discussed, but my general feeling is that the user should be able to choose which part of the plot to display during the simulation, while previously the bounds were just automatically updated.
- Some performance improvements, caused by unnecessary plot update multiple times in a row.

### DIFF
--- a/force_wfmanager/ui/review/plot.py
+++ b/force_wfmanager/ui/review/plot.py
@@ -102,13 +102,20 @@ class BasePlot(BaseDataView):
     # View
     # ----
 
-    view = View(
-        VGroup(
-            HGroup(Item("x"), Item("y")),
-            UItem("_plot", editor=ComponentEditor()),
-            VGroup(UItem("reset_plot", enabled_when="reset_enabled")),
+    axis_hgroup = Instance(HGroup)
+
+    def _axis_hgroup_default(self):
+        return HGroup(Item("x"), Item("y"))
+
+    def default_traits_view(self):
+        view = View(
+            VGroup(
+                self.axis_hgroup,
+                UItem("_plot", editor=ComponentEditor()),
+                VGroup(UItem("reset_plot", enabled_when="reset_enabled")),
+            )
         )
-    )
+        return view
 
     # --------------------
     # Defaults and getters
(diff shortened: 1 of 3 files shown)


Corwinpro:
> LGTM, thanks for this - it makes the code much simpler and easier to refactor.
> 
> Regarding the 'to do' rescaling, this behaviour seems fine for the time being. We can discuss what the ideal behaviour should be and try to implement it without causing too much confusion in the listeners.

Sorry I didn't update the message. The todo is now fixed.

Corwinpro:
@flongford I will merge the PR if you don't mind. It doesn't go to the `master` branch, so in case something is not quite there yet, we can quickly fix it.

Thank you!

flongford:
> @flongford I will merge the PR if you don't mind. It doesn't go to the `master` branch, so in case something is not quite there yet, we can quickly fix it.
> 
> Thank you!

Great, please go ahead

PR #362, opened 2020-02-21
DEV: Delivering UIEventMixin events

### LINKED ISSUE #361: All events are delivered to the WfManager
All events (regardless of their purpose) are dumped and delivered to the WfManager via the ZMQ socket. This is a potentially unsafe behaviour that will result in the ZMQ Poller being stuck if there are too many irrelevant events in the queue.

Corwinpro:
This PR closes #361 

### Summary

We introduced a small `UIEventMixin` class in [#290](https://github.com/force-h2020/force-bdss/pull/290) that indicates that this particular event type should be parsed and sent to the WorkflowManager.

Here we add the actual logic to filter this event type before delivering via the `UINotification.deliver`

### Done

- Added a `if isinstance(event, UIEventMixin)` check in the `UIEventMixin.deliver` method. Only events that are subclasses of the `UIEventMixin` will be dumped and sent to the ZMQ server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
     - if [[ ${TRAVIS_OS_NAME} == "linux" ]] ; then export PATH=${HOME}/edm/bin:${PATH} ; fi
     - if [[ ${TRAVIS_OS_NAME} == "osx" ]]; then export PATH="${PATH}:/usr/local/bin" ; fi
     - edm install --version 3.6 -y click setuptools
-    - git clone git://github.com/force-h2020/force-bdss.git
+    - git clone git://github.com/force-h2020/force-bdss.git -b dev/data-source-event
     - pushd force-bdss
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ before_install:
     - if [[ ${TRAVIS_OS_NAME} == "linux" ]] ; then export PATH=${HOME}/edm/bin:${PATH} ; fi
     - if [[ ${TRAVIS_OS_NAME} == "osx" ]]; then export PATH="${PATH}:/usr/local/bin" ; fi
     - edm install --version 3.6 -y click setuptools
-    - git clone git://github.com/force-h2020/force-bdss.git
+    - git clone git://github.com/force-h2020/force-bdss.git -b dev/data-source-event
```
flongford: GitHub branch reminder

PR #365, opened 2020-02-26
ENH: Refactoring AnalysisModel

Corwinpro:
### Summary

This PR attempts to refactor the `AnalysisModel` and `BaseDataView`.
- Remove the explicit `MCOEvent` type dependency in the `WfManagerSetupTask`
- Allow `UIEventMixin` events to be received by the `AnalysisModel` _during_ the `Workflow` execution
- Simplify the `BaseDataView` and `Plot` classes by removing unnecessary dependencies and attributes.

### Done
- Implemented a new `AnalysisModel` class. Data is added to the `AnalysisModel` row by row; data can be added to the current row by column label, or inserted to the current row as a chunk.  
- We try to retain the basic API the AnalysisModel class has, but some of the used data structures are now different.
  - `value_names` changed to `header`
  - `selected_step_indices` temporarily removed from the `AnalysisModel`, as it should be revised
  - `notify`, `clear` and `clear_steps` are the public method to update and reset the `AnalysisModel`
  - Data that is generated at runtime can be added to the last row of the `AnalysisModel` with `AnalysisModel.notify({"column_name": "value"}). The key value (`"column_name"` in this case) corresponds to the column name of the model (which is an element of the `header`).
- If `AnalysisModel` with an empty header is notified with some data, this data is used as the header.
- `BaseDataView.data_array` attribute has been removed.
  - The `data_array` duplicated the `AnalysisModel.evaluation_steps`, and it is undesirable to store the same data twice.
  - The `AnalysisModel` now implements a `.column(str)` or `.column(int)` method. We can access the specific _column_ of the `AnalysisModel` using this method provided the numerical index of the column, or the column name.


force_wfmanager/ui/review/plot.py
```diff
@@ -288,17 +288,16 @@ def _update_plot_x_axis(self):
...
         This method is called when the `x` axis is changed.
         """
-        if self.x == "" or len(self.data_arrays) == 0:
+        if self.x == "" or self.analysis_model.is_empty:
             self._plot_data.set_data("x", [])
         else:
             self._plot.x_axis.title = self.x
-            x_index = self.analysis_model.value_names.index(self.x)
-            self._plot_data.set_data("x", self.data_arrays[x_index])
+            self._plot_data.set_data("x", self.analysis_model.column(self.x))
```
flongford: So presumably removing the `on_trait_change` handlers from both `_update_plot_x_axis` and `_update_plot_y_axis` methods, in exchange for:

```
@on_trait_change("x,y")
def _update_plot_axis(self):
    self._update_plot_x_axis()
    self._update_plot_y_axis()
```

would mean that the number of data points displayed should be consistent between axis (as long as the `analysis_model` is not updated between calls to obtain the x and y data).

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -453,10 +452,13 @@ def _server_event_mainthread(self, event):
...
-            self.analysis_model.value_names = tuple(event.serialize())
             self.computation_running = True
-        elif isinstance(event, MCOProgressEvent):
-            self.analysis_model.add_evaluation_step(event.serialize())
+        try:
+            event_data = event.serialize()
+        except AttributeError:
+            return
+
+        self.analysis_model.notify(event_data)
```
flongford: At the moment, could the `analysis_model.notify` method handle any data that didn't come from a `MCOProgressEvent`? If not, then we should probably keep this dependency hardcoded here, and aim to replace it with a more generic events handler in the future.

Corwinpro: The `notify` method is a way to insert data into the current row of the `AnalysisModel`. The `analysis_model.notify` can handle two formats:
- `{column_label: data_to_add}`
- `(list / tuple of data_to_add)`
The `AnalysisModel` knows nothing about events and their types.

force_wfmanager/ui/review/plot.py
Corwinpro: Yes, that would solve the problem, thank you. My only concern here is that we would need to use `@on_trait_change("x,y,color_by")` for the color plot, which essentially is a bad design of handling multiple axis in general.

force_wfmanager/wfmanager_setup_task.py
Corwinpro: Followed by the offline discussion, it is clear that implementing an event handler is very important here.
The main point is that the `AnalysisModel` should somehow perform "data filtering" before storing the data.
A possible solution would be to 
- Move the implementation of the `AnalysisModel` to the BDSS;
- Continue to instantiate the `AnalysisModel` in the workflow manager;
- Implement `MCOEvent`s preprocessing as a part of the `AnalysisModel.notify`.

force_wfmanager/ui/review/plot.py
Corwinpro: Shall we leave this for future PRs? This is definitely a UI problem that has to be resolved. At the same time, it has a predictable behaviour, and doesn't break things badly.

flongford: Agreed - let's raise an issue for this as it is a small UX 'bug'

Corwinpro: Opened an issue #367

force_wfmanager/wfmanager_setup_task.py
flongford: This has been resolved for the time being by keeping the hard coded filtering of event types in the `WfManagerSetupTask` so that the data handling is secure, whilst commenting that we should move the logic onto the `AnalysisModel` in future.

PR #370, opened 2020-03-03
ENH: Control enabling of SetupPane UI features during BDSS run

### LINKED ISSUE #358: Setup view: Workflow configuration blocked when optimization running
When we `"Run"` the Workflow, it is not possible to inspect what are the parameters of the Workflow. This appears to be a undesirable behaviour from the UX point of view, as the user can't know the exact setup of the experiment they are running.

A possible solution would be to display the workflow setup on the right hand side of the setup pane, replacing the "Available Plugins" section (which is more or less useless).

flongford:
This PR closes #358 

It adds a new attribute `ui_enabled` to the `SetupPane` class, which controls whether the UI features are enabled or not during a BDSS run. By doing so we are also able to remove the same control on the `SidePane` tree view, so that it remains selectable during run time and allows the user to look at the `Workflow` state without being able to edit anything.

The `SidePane.ui_enabled` attribute is retained, as it is involved in the logic defining the status of the run buttons.

### Future Work
- The view objects of the `SetupPane` are a bit convoluted, and we are not able to simply set the enable status of a single object. It would be great if we could refactor the view into something that has a single `Group`, so that `visible` or `enabled` states could be easily set for the entire pane.

### DIFF
--- a/force_wfmanager/tests/test_wfmanager_setup_task.py
+++ b/force_wfmanager/tests/test_wfmanager_setup_task.py
@@ -331,6 +331,7 @@ class TestWFManagerTasks(GuiTestAssistant, TestCase):
             hook_manager = self.setup_task.ui_hooks_managers[0]
 
             self.assertTrue(self.setup_task.side_pane.ui_enabled)
+            self.assertTrue(self.setup_task.setup_pane.ui_enabled)
             self.assertFalse(hook_manager.before_execution_called)
             self.assertFalse(hook_manager.after_execution_called)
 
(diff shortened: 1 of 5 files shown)


flongford:
Before merging, I just want to test the UI response speed when it locks up the file tree - we should make a note if this introduces some delay in the 'run' command

flongford:
> Before merging, I just want to test the UI response speed when it locks up the file tree - we should make a note if this introduces some delay in the 'run' command

Update: UI reponse speed seems normal

PR #373, opened 2020-04-02
switch if length of data set rather than data

### LINKED ISSUE #369: Plot axis bounds for repeating values
The [calculate_axis_bounds](https://github.com/force-h2020/force-wfmanager/blob/9f070c8b7ec279b89dac6f78c1aec717ae3f243e/force_wfmanager/ui/review/plot.py#L359) method does not properly handle the case when the `len(data) > 1` but the actual data values are the same. 

Consider the case `data = [1,1,1]`: the bounds must be calculated as
```
            bounds = (data[0] - 0.5, data[0] + 0.5)
```
While currently the first branch `if len(data) > 1:` is executed.

To fix this bug, the `len(set(data))` should be used instead of `len(data)`.

sparsonslab:
Closes #369. Length of set is calculated before if-else ladder, as otherwise set would have to be called twice on a long list with a set size of 1.

### DIFF
--- a/force_wfmanager/ui/review/tests/test_plot.py
+++ b/force_wfmanager/ui/review/tests/test_plot.py
@@ -230,7 +230,7 @@ class TestBasePlot(GuiTestAssistant, unittest.TestCase, UnittestTools):
             mock_update.assert_called()
 
         self.plot._update_plot()
-        self.assertEqual((1.8, 4.2, 3.0, 3.0), self.plot._get_plot_range())
+        self.assertEqual((1.8, 4.2, 2.5, 3.5), self.plot._get_plot_range())
 
     def test_remove_value_names(self):
         self.analysis_model.header = ("density", "pressure")
--- a/force_wfmanager/ui/review/plot.py
+++ b/force_wfmanager/ui/review/plot.py
@@ -359,7 +359,7 @@ class BasePlot(BaseDataView):
     def calculate_axis_bounds(data):
         if len(data) < 1:
             bounds = (-1, 1)
-        if (len(data) > 1) and (data[0] != data[1]):
+        elif (len(data) > 1) and (data[0] != data[1]):
                 axis_max = max(data) * 1.0
                 axis_min = min(data)
                 axis_spread = abs(axis_max - axis_min)
@@ -368,7 +368,6 @@ class BasePlot(BaseDataView):
                 bounds = (axis_min, axis_max)
         else:
             bounds = (data[0] - 0.5, data[0] + 0.5)
-
         return bounds
 
     def _reset_zoomtool(self):
--- a/force_wfmanager/ui/review/plot.py
+++ b/force_wfmanager/ui/review/plot.py
@@ -357,17 +357,18 @@ class BasePlot(BaseDataView):
 
     @staticmethod
     def calculate_axis_bounds(data):
-        if len(data) < 1:
-            bounds = (-1, 1)
-        elif (len(data) > 1) and (data[0] != data[1]):
+        set_length = len(set(data))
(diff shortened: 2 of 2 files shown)


flongford:
> ... excuse me forgot the commit comment. Now we don't need to use the set operation. Just check the equality of the first two elements (if len > 1)

I'm not sure this would still work if other elements in the list had the possessed the same value. Is there any preference for not using `set` here?

sparsonslab:
> > ... excuse me forgot the commit comment. Now we don't need to use the set operation. Just check the equality of the first two elements (if len > 1)
> 
> I'm not sure this would still work if other elements in the list had the possessed the same value. Is there any preference for not using `set` here?

Because set() might take a long time on a large list. I know these new-fangled computers are very fast, but its still neat to act like we need to be efficient. 
note, this wouldn't make any difference if len() was o(n) as well, but it is actually o(1) as it just returns an attribute of the list.

flongford:
> > > ... excuse me forgot the commit comment. Now we don't need to use the set operation. Just check the equality of the first two elements (if len > 1)
> > 
> > 
> > I'm not sure this would still work if other elements in the list had the possessed the same value. Is there any preference for not using `set` here?
> 
> I think I see.

Now I'm questioning my logic - hang on, you might have been right previously....

sparsonslab:
> > > > ... excuse me forgot the commit comment. Now we don't need to use the set operation. Just check the equality of the first two elements (if len > 1)
> > > 
> > > 
> > > I'm not sure this would still work if other elements in the list had the possessed the same value. Is there any preference for not using `set` here?
> > 
> > 
> > I think I see.
> 
> Now I'm questioning my logic - hang on, you might have been right previously....

No, you're right. My code would have failed with e.g. data = [1, 1, 0]

flongford:
> > > > > ... excuse me forgot the commit comment. Now we don't need to use the set operation. Just check the equality of the first two elements (if len > 1)
> > > > 
> > > > 
> > > > I'm not sure this would still work if other elements in the list had the possessed the same value. Is there any preference for not using `set` here?
> > > 
> > > 
> > > I think I see.
> > 
> > 
> > Now I'm questioning my logic - hang on, you might have been right previously....
> 
> No, you're right. My code would have failed with e.g. data = [1, 1, 0]

Right, yes that makes sense - phew, got there in the end

force_wfmanager/ui/review/tests/test_plot.py
```diff
@@ -230,7 +230,7 @@ def test_select_plot_axis(self):
             mock_update.assert_called()
 
         self.plot._update_plot()
-        self.assertEqual((1.8, 4.2, 3.0, 3.0), self.plot._get_plot_range())
+        self.assertEqual((1.8, 4.2, 2.5, 3.5), self.plot._get_plot_range())
```
flongford, line 233: Just checking - presumably this test was incorrect before? 

It looks like the updated version should be what we expect to return

flongford:
Ah, one other addition that we should include is a unit test of `calculate_axis_bounds ` for the edge case where `data = [1, 1, 1]` (to show the bug is fixed)

force_wfmanager/ui/review/tests/test_plot.py
sparsonslab: if you do print(data) at the start of calculate_axis_bounds(data), you'll find
[2. 4.]   > 1.8, 4.2
[3. 3.]  > 2.5, 3.5  (not 3.0, 3.0)

So yes the original assert was wrong.

flongford: I agree - looking closer I think this change also provides the unit test that I requested

PR #374, opened 2020-04-06
Common toolbar

### LINKED ISSUE #371: ToolBar items duplicated in both WfManagerReviewTask and WfManagerSetupTask
Currently the tool bar in both review and setup task contains duplicated code, which needs to be updated whenever any objects change state during run time.

Using the envisage `TasksExtensions` API to only generate one instance of each unique element in the tool bar would be desirable, and simplify the icon state updates (enabled/disabled during an MCO run).

https://docs.enthought.com/envisage/tasks_user_manual/extensibility.html

sparsonslab:
Will close #371.

Common menu- and tool-bar items are now added to the setup and review tasks as a global extension.

The methods called by each global TaskAction are still all methods of the setup task. As each TaskAction "method=" argument has to refer to a common string description of the method path, this requires that the setup task has a reference to itself. The alternative would be to put these methods in another class, but I don't want to break stuff too-much.

No doubt this won't pass any tests, but we can do that next.

The UI does work on my machine.


force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,132 @@
...
+    def run_tools(self):
+        """ Run/pause/stop workflow tools schema.
+        """
+        return SToolBar(
+                TaskAction(
+                    name="Run",
+                    tooltip="Run Workflow",
+                    image=ImageResource("baseline_play_arrow_black_48dp"),
+                    method="setup_task.run_bdss",
+                    image_size=(64, 64),
```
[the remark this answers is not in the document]
sparsonslab: Doing that blocks those tool buttons (run/stop/pause) completely. I couldn't work out why. But I think it only makes a difference to the appearance of the tool buttons. e.g. stop won't do anything if nothing is running.

force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,132 @@
...
+            ]
+
+    def file_menu(self):
+        """ File menu schema.
+        """
+        return SMenu(
+                TaskAction(
+                    name="Open Workflow...",
+                    method="setup_task.open_workflow",
+                    enabled_name="save_load_enabled",
```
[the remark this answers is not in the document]
sparsonslab: These enabled_name (for the menu bar) can also be removed - at the moment they don't function. If they did, then  (like for the tool bar) they would totally shut off the options.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -122,6 +122,12 @@ class WfManagerSetupTask(Task):
     #: Review Task
     review_task = Instance(Task)
 
+    #: Setup Task: this will be a self-reference
+    # that allows the TaskActions added by the global task extension
+    # to refer to methods defined by the setup task instance,
+    # whether that extension is added to setup or review.
+    setup_task = Instance(Task)
```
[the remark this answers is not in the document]
sparsonslab: I'm beginning to think this is not a reference to itself, but another instance of WfManagerSetupTask. That would explain why the enabled_name calls don't work - they refer to that other instance where the state is not changing.

force_wfmanager/wfmanager_global_task.py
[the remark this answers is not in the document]
flongford: What about when you try starting the `force_wfmanager` with a complete workflow? The Run button should be blocked out if the workflow being built is not valid:

https://github.com/force-h2020/force-wfmanager/blob/master/force_wfmanager/ui/setup/side_pane.py#L96

force_wfmanager/wfmanager_setup_task.py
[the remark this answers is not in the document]
sparsonslab: ... no it is a self-reference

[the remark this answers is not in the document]
flongford: I think this is working correctly - I have a commit that demonstrates how to get the `enabled_name` references working properly for the start / stop / pause buttons:

https://github.com/force-h2020/force-wfmanager/commit/c74269ee542414a10931811bf92fdd72aa5c6c26

sparsonslab:
I can't get the state changes to work for the run/pause/stop as I have no way of now accessing them the tasks. Usually,

line 525, wfmanager_setup_task
self.tool_bars[0].items[2],
self.review_task.tool_bars[0].items[2],

But now the extended run/pause/stop toolbar is not an item in self.tool_bars[].

sparsonslab:
... but all the state change seems to do is change the name and tooltip stuff, so maybe we don't need this?

flongford:
> I can't get the state changes to work for the run/pause/stop as I have no way of now accessing them the tasks. Usually,
> 
> line 525, wfmanager_setup_task
> self.tool_bars[0].items[2],
> self.review_task.tool_bars[0].items[2],
> 
> But now the extended run/pause/stop toolbar is not an item in self.tool_bars[].

Yes, I see the problem - I think we could probably move this method to the `WfManagerGlobalTask` class, since it is just a workaround to change the pause icon.

sparsonslab:
> > I can't get the state changes to work for the run/pause/stop as I have no way of now accessing them the tasks. Usually,
> > line 525, wfmanager_setup_task
> > self.tool_bars[0].items[2],
> > self.review_task.tool_bars[0].items[2],
> > But now the extended run/pause/stop toolbar is not an item in self.tool_bars[].
> 
> Yes, I see the problem - I think we could probably move this method to the `WfManagerGlobalTask` class, since it is just a workaround to change the pause icon.

... ah but the task extention isn't itself meant to have a reference to the toolbar, it just passes the toolbar on as a schema.  Also the state information (needed) is still tied up in the set-up task.

Of course I add the toolbar as an attribute in the task extension, and then add a reference to this in the set-up and review tasks, but then this would completly go against the philosophy of the task extension.

I think Corran would know the "elegant" answer to this. There must be a way to get a reference to the toolbar added by the global task extension, from any task.

flongford:
> ... ah but the task extention isn't itself meant to have a reference to the toolbar, it just passes the toolbar on as a schema. Also the state information (needed) is still tied up in the set-up task.

Very true. Hmmmm... this part requires a bit more thinking to resolve. Are you able to avoid it for the time being? Just removing the `@on_trait_change` decorator will prevent the method from being called - I think it is probably the only code that is invoking `self.tool_bars`

> Of course I add the toolbar as an attribute in the task extension, and then add a reference to this in the set-up and review tasks, but then this would completly go against the philosophy of the task extension.
> 
> I think Corran would know the "elegant" answer to this. There must be a way to get a reference to the toolbar added by the global task extension, from any task.

Agreed - this would be a good discussion to have with him next week.

flongford:
Update:

> I can't get the state changes to work for the run/pause/stop as I have no way of now accessing them the tasks. Usually,
> 
> line 525, wfmanager_setup_task
> ```python
> self.tool_bars[0].items[2],
> self.review_task.tool_bars[0].items[2],
> ```
> But now the extended run/pause/stop toolbar is not an item in self.tool_bars[] or self.review_task.tool_bars[].

We could achieve the same result (changing icon / name on a tool bar object) by having 2 toolbar objects that become visible / not visible in reaction to the same event.

This is less elegant, but is robust (if the underlying logic controlling the `visible_when` attributes are) and can be easily changed in the future if the UI requirements also change.

Something like:

```python
TaskAction(
    name="Pause",
    tooltip="Pause Workflow",
    method="setup_task.pause_bdss",
    enabled_name='setup_task.computation_running',
    visible_when='setup_task._not_paused',
    image=ImageResource("baseline_pause_black_18dp"),
    image_size=(64, 64),
),
TaskAction(
    name="Resume",
    tooltip="Resume Workflow",
    method="setup_task.pause_bdss",
    enabled_name='setup_task.computation_running',
    visible_when='setup_task._paused'
    image=ImageResource("baseline_skip_next_black_18dp.png"),
    image_size=(64, 64),
)
```

With the undying logic controlling `setup_task._paused` and `setup_task._not_paused` is defined in the `WfManagerSetupTask` class

sparsonslab:
There seemed to be a redundant method  `WfManagerSetupTask._pause_fired()`: this (and one of the tests) was the only method that accessed the toolbar list (the problem), but I didn't realize it was redundant and so could just be taken out.  With this out everything works.

flongford:
> There seemed to be a redundant method `WfManagerSetupTask._pause_fired()`: this (and one of the tests) was the only method that accessed the toolbar list (the problem), but I didn't realize it was redundant and so could just be taken out. With this out everything works.

Yes, this is the problematic method but it is currently controlling the `Pause --> Resume` toolbar icon appearance change, so getting rid of it will also remove this feature. Did you try the suggested work around in order to keep the feature (using the `visible_name` options)?

sparsonslab:
> > There seemed to be a redundant method `WfManagerSetupTask._pause_fired()`: this (and one of the tests) was the only method that accessed the toolbar list (the problem), but I didn't realize it was redundant and so could just be taken out. With this out everything works.
> 
> Yes, this is the problematic method but it is currently controlling the `Pause --> Resume` toolbar icon appearance change, so getting rid of it will also remove this feature. Did you try the suggested work around in order to keep the feature (using the `visible_name` options)?

When tried your work-around, but couldn't get the toolbar elements to actually appear. I couldn't figure out which combination of `enabled_name` and `visible_name` would solve this.

force_wfmanager/wfmanager_global_task.py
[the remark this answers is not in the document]
flongford: I think trying to set these `save_load_enabled` references to `setup_task.save_load_enabled` should now work

force_wfmanager/wfmanager_review_task.py
```diff
@@ -45,9 +44,6 @@ class WfManagerReviewTask(Task):
     #: and table
     analysis_model = Instance(AnalysisModel, allow_none=False)
 
-    #: Is the 'run' toolbar button active
-    run_enabled = Bool(True)
-
     #: Are the saving and loading menu/toolbar buttons active
     save_load_enabled = Bool(True)
```
flongford, line 48: I think this attribute `save_load_enabled` can now be removed, along with the `sync_save_load_enabled` method on line 305

force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,147 @@
...
+                ),
+                TaskAction(
+                    name="Save Workflow as...",
+                    method="setup_task.save_workflow_as",
+                    enabled_name="setup_task.save_load_enabled",
+                    accelerator="Shift+Ctrl+S",
+                ),
+                TaskAction(name="Plugins...",
+                           method="setup_task.open_plugins"),
+                TaskAction(name="Exit", method="exit"),
```
corranwebster, line 85: Names are important for Qt and MacOS: you want to vary the name depending on the platform you are on.  SOmething like:
```
import platform

IS_WINDOWS = (platform.system() == "Windows")

class ExitAction(TaskAction):
    """ A standard 'Quit' or 'Exit' menu action. """

    accelerator = "Alt+F4" if IS_WINDOWS else "Ctrl+Q"
    method = "exit"

    def _name_default(self):
        return (u"Exit" if IS_WINDOWS else u"Quit ") + self.application.name
```
This should behave as expected on most platforms, including not getting two "Exit"/"Quit" menus (On OS X the quit menu will automatically be shifted from the File menu to the correct place).

force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,147 @@
...
+                # NOTE: Setting id='File' here will automatically create
```
corranwebster, line 86: This should be an issue, not a Note.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -122,6 +127,12 @@ class WfManagerSetupTask(Task):
     #: Review Task
     review_task = Instance(Task)
 
+    #: Setup Task: this will be a self-reference
+    # that allows the TaskActions added by the global task extension
```
corranwebster, line 131: Nit: you want all of these comments to be `#:` for Sphinx autodoc to pick it up.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -303,6 +250,11 @@ def _zmq_server_default(self):
     #     Listeners
     # ------------------
 
+    def _get__not_paused(self):
```
corranwebster, line 253: Could make this a `@cached_property`; possibly not worth the effort.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -343,7 +295,8 @@ def sync_review_task(self):
             for task in self.window.tasks:
                 if task.name == "Review":
                     self.review_task = task
-                    self.review_task.run_enabled = self.run_enabled
+                elif task.name == "Workflow Setup":
```
corranwebster, line 298: I realize that this is duplicating the approach of existing code, but in this case I would recommend using the `task.id` for both tests rather than the `task.name`.

Don't need to fix in this PR if the problem occurs in many places, but if you don't fix it, open an issue.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -545,8 +633,10 @@ def load_workflow(self, file_path):
                     "Project file found instead of Workflow "
                     "file during start up.",
                     informative="Analysis data will not be "
-                    "loaded into WfManager upon launch. You can load a "
-                    "Project file using 'Open Project' in the Review Task.",
+                                "loaded into WfManager upon launch. "
```
corranwebster, line 636: Style nit: I'd put parentheses around multi-line strings of this form:
```
                    informative=(
                        "Analysis data will not be "
                        ...
                    )
```

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -505,7 +458,142 @@ def prepare_destroy(self):
...
+        dlg = PluginDialog(plugins)
+        dlg.edit_traits()
+
+    # Menu/Toolbar Methods
+    def switch_task(self):
+        """Switches to the review task and verifies startup setting are
+        correct for toolbars/menus etc."""
+        self.window.activate_task(self.review_task)
+
+    def exit(self):
```
corranwebster, line 553: I realize that this is copied code, but `exit` is an application-level thing rather than a Task-level thing, conceptually.

The default exit behaviour for Tasks applications is here: https://github.com/enthought/envisage/blob/master/envisage/ui/tasks/tasks_application.py#L273-L298

This should do exactly what you want, so it is this method that you want to call, rather than this one.  There is a `TasksApplicationAction` that is suited for this use.

force_wfmanager/wfmanager_global_task.py
corranwebster: As noted below, this should probably be a `TasksApplicationAction` which calls the application's `exit` method.

corranwebster: And, now that I think about it, probably shouldn't even be in _this_ extension, and may not be needed at all.

force_wfmanager/tests/test_wfmanager.py
```diff
@@ -138,14 +138,14 @@ def test_result_task_exit(self):
             for window in self.wfmanager.windows:
                 window.close = mock.Mock(return_value=True)
             self.review_task.exit()
-            self.assertTrue(self.review_task.window.close.called)
+            self.assertTrue(self.wfmanager.exit)
```
sparsonslab, line 141: The tests passed, but I got some funny message:

`KeyError: <weakref at 0x0000023A93E84EF8; to 'TasksPlugin' at 0x0000023A93CAD938>`

force_wfmanager/wfmanager_global_task.py
```diff
@@ -12,6 +13,18 @@
...
+
+class ExitAction(TaskAction):
+    """ A standard 'Quit' (Mac) or 'Exit' (Windows) menu action. """
+
+    accelerator = "Alt+F4" if IS_WINDOWS else "Ctrl+Q"
+    method = "setup_task.exit"
+
+    def _name_default(self):
+        return u"Exit" if IS_WINDOWS else u"Quit"
+
```
sparsonslab, line 27: Couldn't find much doc on ApplicationTask, so couldn't figure out how to implement that. But `set_task.exit` now calls the application's exit method.

force_wfmanager/wfmanager_setup_task.py
sparsonslab: Couldn't find much doc on `ApplicationTask`, so couldn't figure out how to implement that. But `set_task.exit` now calls the application's exit method.

force_wfmanager/wfmanager_setup_task.py
```diff
@@ -135,83 +146,19 @@ class WfManagerSetupTask(Task):
...
                     tooltip="View Results",
                     image=ImageResource("baseline_bar_chart_black_48dp"),
                     method="switch_task",
                     image_size=(64, 64),
-                ),
+                    id='View Results'
+                ), id='running'
             ),
             SToolBar(
                 TaskAction(
```
flongford, line 164: Some of these `TaskAction` items can be removed as they are now already defined in `WfManagerGlobalTask`

force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,156 @@
...
+
+class ExitAction(TaskAction):
+    """ A standard 'Quit' (Mac) or 'Exit' (Windows) menu action. """
+
+    accelerator = "Alt+F4" if IS_WINDOWS else "Ctrl+Q"
+    method = "setup_task.exit"
+
+    def _name_default(self):
+        return u"Exit" if IS_WINDOWS else u"Quit"
```
flongford, line 26: I don't think this class is working properly - it's not appearing in the `File` drop down menu. Some simple unit tests might help to see whether the problem lies with the class definition, or where we are instantiating it in line 98.

force_wfmanager/wfmanager_global_task.py
```diff
@@ -0,0 +1,156 @@
...
+log = logging.getLogger(__name__)
+
+IS_WINDOWS = (platform.system() == "Windows")
+
+
+class ExitAction(TaskAction):
+    """ A standard 'Quit' (Mac) or 'Exit' (Windows) menu action. """
+
+    accelerator = "Alt+F4" if IS_WINDOWS else "Ctrl+Q"
+    method = "setup_task.exit"
```
flongford, line 23: I think Corran's suggestion meant to move this method reference to the `TaskApplication` object (`WfManager`, rather than `WfManagerSetupTask`).

This does not seem straight forward, so it is something probably best dealt with in an additional PR

force_wfmanager/wfmanager_setup_task.py
flongford: I think we should deal with this in a new issue / PR, since it lies a bit out of scope of the current work

force_wfmanager/tests/test_wfmanager.py
flongford: This is failing on the Travis build, so we'll need to work out what's going on here. I think we can do this in a new PR, since it seems like it will need a bit of attention.

force_wfmanager/wfmanager_global_task.py
flongford: It looks like the exit routine has been overridden in a `TaskWindow` subclass in order to provide a custom message / save function: https://github.com/force-h2020/force-wfmanager/blob/master/force_wfmanager/wfmanager.py#L67

This might require a bit of work to untangle

corranwebster:
Definitely do the close stuff in a different PR if it is at all complex (which it sounds like it might be).

force_wfmanager/wfmanager_global_task.py
corranwebster: It shouldn't on Mac OS, it should appear in the main application menu (probably called "Python" when you typically run).  See https://doc.qt.io/qt-5/macos-issues.html#menu-bar

flongford: Ah, I see - then I think the master branch had overridden this feature to provide an extra item in the `File` drop down:

Screenshot 2020-04-20 at 16 48 57

But this no longer appears in the current branch:

Screenshot 2020-04-20 at 16 50 56

From the comments above, it seems that the current behaviour (exit only through `Python` from down) should be put in place, but addressed in a new PR.

sparsonslab:
Glad it wasn't just me then. Shall I just revert this branch to the previous commit then?

flongford:
> Glad it wasn't just me then. Shall I just revert this branch to the previous commit then?

You can, yes that might be easier - just make sure you then re-implement any other changes you made during that commit.

sparsonslab:
I have taken out exit-menu-related stuff in the global task. The set-up task exit function is restored to its previous state, but I suppose that will be removed in your new PR.

force_wfmanager/wfmanager_global_task.py
flongford: See #378

force_wfmanager/wfmanager_setup_task.py
flongford: @sparsonslab - did you check this out? I think the comment got lost amongst the sea of application exit talk

PR #375, opened 2020-04-15
ENH: Suggestion to get switch between 'Pause' and 'Resume' icons working

flongford:
Demonstration of using `visible_when` settings on `TaskAction` objects in order to switch between 'Pause' and 'Resume' display appearances.

### Feature:
- Before an MCO run, the 'Pause' button appears not enabled
- During an MCO run, if the 'Pause' button is clicked, it should change appearance to the 'Resume' button
- If the 'Resume' button is clicked during a pause, the MCO will resume and the 'Pause' button will be re-displayed
- If the 'Stop' button is clicked during a pause, the MCO will stop and the 'Pause' button will be re-displayed and not enabled.


flongford:
> Maybe you can explain what did the trick this afternoon. Was it something to do with the run_enabled methods/traits being removed in setup task?

Yes, there was an issue with duplicating the `run_enabled` attribute, since it was being reassigned each time the tasks were changed by the `WfManagerSetupTask.sync_review_task` method. Additionally, since we now only use `setup_task.run_enabled` as a reference in the toolbar, this duplication was unnecessary.

Issue #381, opened 2020-04-24
Parameters and KPIs lost when changing MCO


sparsonslab:
When a new MCO is loaded in the set-up pane/task, the workflow's parameters and KPI's vanish and have to be reset from scratch. This is a problem if the user wants to compare the performance of different MCOs on the same workflow.

PR #383, opened 2020-05-06
Revert params

sparsonslab:
## Description

When the user selects a new MCO, the parameters and kpis are wiped and then the user has to
set them again. (#381). This makes a small change to `WorkflowTree.new_mco()`, so that the new MCO's parameters and kpis are set to those of the old MCO.

### Related Issues
#381

## Changes
This makes a small change to `WorkflowTree.new_mco()`, that sets the new MCO's parameters and kpis to those of the old MCO. The kpis can just be copied as they are self-contained. Each parameter refers to the MCOModel and plugin (via their factories) and these are changed appropriately.

### DIFF
--- a/force_wfmanager/ui/setup/tests/test_workflow_tree.py
+++ b/force_wfmanager/ui/setup/tests/test_workflow_tree.py
@@ -192,7 +192,7 @@ class TestWorkflowTree(WfManagerBaseTestCase):
 
         # have ONLY two parameters been transferred?
         # (the third categorical parameter of the old_model
-        # should not be transferred as it is not allowed by the new model).
+        # should not be filtered out as it is not allowed by the new model).
         self.assertEqual(2, len(new_model.parameters))
         for p in new_model.parameters:
             self.assertNotIsInstance(p, CategoricalMCOParameter)
--- a/force_wfmanager/ui/setup/tests/test_workflow_tree.py
+++ b/force_wfmanager/ui/setup/tests/test_workflow_tree.py
@@ -147,6 +147,10 @@ class TestWorkflowTree(WfManagerBaseTestCase):
 
         # create old model
         old_factory = DummyMCOFactory(plugin={'id': 'bcy356', 'name': 'old'})
+        old_factory.parameter_factory_classes = [
+            RangedVectorMCOParameterFactory,
+            RangedMCOParameterFactory
+        ]
         old_model = BaseMCOModel(factory=old_factory)
         old_model.parameters = [
             RangedMCOParameter(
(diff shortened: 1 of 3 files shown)


force_wfmanager/ui/setup/workflow_tree.py
```diff
@@ -646,7 +646,37 @@ def new_layer(self, ui_info, object):
...
+            object.model.mco_model = self.system_state.entity_creator.model
+
+            # Set the new model's parameters and kpis to that of the old.
+            # Each parameter's factory attribute should refer to a new
+            # instance of the relevant MCOParameterFactory, that references
+            # the appropriate MCOFactory, plugin-id and plugin-name.
+            # This is done by calling MCOParameterFactory(<new MCOFactory>).
+            params_new = []
+            for p in params:
+                p.factory = p.factory.__class__(object.model.mco_model.factory)
```
flongford, line 670: Note: We should try not refer to protected attributes of a class externally - if there is a missing functionality then it's best to introduce this into the class itself and provide a public method to call. 

Additionally, I think we need to check here whether the each of the old `BaseMCOParameterFactory` classes are supported by the new `BaseMCOFactory` class (`object.model.mco_model.factory`). 

Putting both these together, we could use the `BaseMCOFactory.parameter_factory_classes` attribute to have have something like:

```python
mco_factory = object.model.mco_model.factory

for p in params:
    for factory_cls in mco_factory.parameter_factory_classes:
        if isinstance(p.factory, factory_cls):
            p.factory = factory_cls(mco_factory)
            params_new.append(p)
```

force_wfmanager/ui/setup/tests/test_workflow_tree.py
```diff
@@ -137,6 +143,55 @@ def test_mco_kpis_selected(self):
...
+        old_model.parameters = [
+            RangedMCOParameter(
+                initial_value=1.0,
+                factory=RangedVectorMCOParameterFactory(old_factory)
+            ),
+            RangedVectorMCOParameter(
+                initial_value=[1.0, 1.0],
+                factory=RangedMCOParameterFactory(old_factory)
+            )
+        ]
```
flongford, line 164: Could we add an additional class and parameter here (`CategoricalMCOParameterFactory`/`CategoricalMCOParameter`?) in `old_model.parameter_factory_classes` and `old_model.parameters`, which will not be supported in `new_factory`?

I expect the rest of the tests to run the same in this case - `new_factory.parameters` will only contain the `RangedMCOParameter` and `RangedVectorMCOParameter` instances.

PR #394, opened 2020-07-03
PyQt 5

sparsonslab:
## Description
Update of the dependencies from PyQt 4 to 5.

### Background
Enthought libraries (Traits/etc) will no longer be maintained for PyQt 4 dependence.

## Changes
- `ADDITIONAL_CORE_DEPS` in `ci/main.py` changed so that PyQt5 is used and other dependencies upgraded around this.

### DIFF
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -10,12 +10,12 @@ DEFAULT_PYTHON_VERSION = "3.6"
 PYTHON_VERSIONS = ["3.6"]
 
 ADDITIONAL_CORE_DEPS = [
-    "pyface>=6.1.2-5",
+    "pyface==7.0.0-3",
     "pygments==2.2.0-1",
     "pyqt5==5.14.2-3",
     "sip==4.17-4",
-    "traits>=6.0.0-1",
-    "traitsui==7.0.0-1",
+    "traits==6.1.0-1",
+    "traitsui==7.0.0-2",
     "numpy>=1.15.4-2",
     "chaco>=4.8.0-1",
     "pyzmq==16.0.0-7",
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -10,14 +10,14 @@ DEFAULT_PYTHON_VERSION = "3.6"
 PYTHON_VERSIONS = ["3.6"]
 
 ADDITIONAL_CORE_DEPS = [
-    "pyface==7.0.0-3",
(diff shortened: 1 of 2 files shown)


sparsonslab:
The linux fail come down to some qt package:  "qt.qpa.plugin". If you search on the Enthought slack lots of other people have had the same problem.

ci/__main__.py
```diff
@@ -10,14 +10,14 @@
 PYTHON_VERSIONS = ["3.6"]
 
 ADDITIONAL_CORE_DEPS = [
-    "pyface==7.0.0-3",
+    "pyface>=6.1.2-5",
```
flongford, line 13: I have a feeling we probably want to be using the latest `pyface` version available to avoid issues with Qt5 - is there a way to enforce this or does it clash with other dependencies?

sparsonslab: Clashes (though to be honest I don't completely understand the errors printed). I'm trying one of Mark's suggestions....

flongford: Hmmm... yes I just pulled and tried it locally. I'm not too sure what's going on either, looks like a clash with `envisage`. Unfortunately from the Travis build logs it looks like the workaround also leads to us losing the latest Traits and TraitsUI packages during the `force_wfmanager` installation:

```
685  The following packages will be downgraded:
686
687    envisage  4.9.2-3  -->  4.9.2-2  596     KiB  *
688    pyface    7.0.0-3  -->  7.0.0-2    1.93  MiB  *
689    traits    6.1.0-1  -->  6.0.0-1  737     KiB  *
690    traitsui  7.0.0-2  -->  7.0.0-1    1.51  MiB  *
```

flongford: @sparsonslab - try removing `pyqt` from your local `force-py36` environment first (`edm remove pyqt`) before installing `force_wfmanager`. I was able to get it working using:

```python
ADDITIONAL_CORE_DEPS = [
    "pyface==7.0.0-3",
    "pygments==2.2.0-1",
    "pyqt5==5.14.2-3",
    "sip==4.17-4",
    "traits==6.1.0-1",
    "traitsui==7.0.0-2",
    "numpy>=1.15.4-2",
    "chaco>=4.8.0-1",
    "pyzmq==16.0.0-7",
    "mock==2.0.0-3"
]
```

Issue #359, opened 2020-02-21
BDSS Stop / Pause icons have delayed response in UI


flongford:
It is possible for the BDSS to take a while to stop or pause a run when requested by the UI, if it is in the middle of a long calculation. Therefore it will appear to the user as if nothing has happened and the UI is unresponsive.

We should come up with an agreed UX procedure that makes sure the user feels that a stop / pause message has been received and the BDSS is waiting to act upon the message. This should also not give the impression that the UI has frozen.

PR #396, opened 2020-07-16
ENH: KPI bounds display based on objective

flongford:
## Description

Displays either upper or lower bounds for KPIs based on objectives.

Upper bounds are only relevant for minimisation objectives lower bounds are only relevant for maximisation objectives.

Target objectives have an upper bound that is a maximum 'distance from target' (could be either above or below raw value).

## Changes
- `KPIModelView` displays upper and lower bounds based on model objective


flongford:
Closing this since https://github.com/force-h2020/force-bdss/pull/359 removes the requirements for KPI target / bounds values. We should instead define the view on the `KPISpecification` class itself and then embed it in the model view (similar to the `MCOParameterModelView` class)

